Issue #13, opened 2019-08-05
Configuration for position of single line statements


dsherret:
For example, deciding on this:

```ts
if (test) statement;
// or
if (test)
    statement;
```

I will probably keep the second as the default because I’m a big fan of it over the first.

https://twitter.com/davidsherret/status/1143491015518621696?s=21

Should be 3 options, one for first, second, then "maintain".

dsherret:
Implemented as `singleBodyPosition`.

PR #19, opened 2019-08-16
Configuration - Single body position (For #13)

dsherret:
This change also adds and makes "whenNotSingleLine" the default for "useBraces".


dsherret:
Code is getting a bit messy, but I love how declarative this is. Makes it super easy to do complex things.

PR #56, opened 2020-01-26
Replace vector parsing with graph

dsherret:
This is faster and formats checker.ts in ~772ms.

1. A graph was being constructed on top of the parsed vectors. This graph generation was pushed up to the parsing stage so only one data structure needed to be created.
2. Resolved condition values are now only stored in the printer if someone gets a reference to a condition, which can then be used in a condition resolver. This saves a lot of time and needless memory usage.
3. In the implementation, UnsafeCell is used instead of RefCell. This is made "safe" by hiding away the use of UnsafeCell.
4. Indent "write items" are now stored with a count value. Previously it was pushing a single indent IR the number of times an indent was necessary.
5. This goes back to having the concept of "Signals" in the parsing, which I am very happy about.
6. Made `Info` and `ConditionReference` implement the `Copy` trait. These objects are extremely lightweight and doing the clones was unnecessary complexity.

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -60,8 +60,6 @@ This is an enum that signals information to the printer.
 * `StartIgnoringIndent` - Signal to the printer that it should stop using indentation.
 * `FinishIgnoringIndent` - Signal to the printer that it should start using indentation again.
 
-In the Rust implementation these exist directly on the `PrintItem` enum.
-
 ## Printer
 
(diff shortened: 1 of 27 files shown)


dsherret:
For some reason I accidentally didn't squash this so the new squashed commit is here on master: https://github.com/dsherret/dprint/commit/2cb6ae4b152c2f128d4dd688d785d48254b04125

Issue #88, opened 2020-02-04
Support top level await


dsherret:
Example:

```ts
async function f() {
    
}
await f();
```

Opened https://github.com/swc-project/swc/issues/626

ry:
It seems https://github.com/swc-project/swc/issues/626 was fixed. 

https://github.com/swc-project/swc/commit/94eac1de89a69ce0f2b30ced7d1c3f9fe48640aa

I guess all that needs to be done is update swc version in dprint to 1.1.18?

dsherret:
@ry waiting on an swc_ecma_parser release: https://crates.io/crates/swc_ecma_parser

But yes, it should work right away after upgrading.

dsherret:
Implemented and released in 0.4.0.

Couple of problems that were resolved:

1. For an expression statement containing an await expression, swc is currently generating an expression statment and an empty statement. This is fixed in swc, but not yet released. I worked around this by getting rid of any expression statements in the output (which is desirable anyway).
2. Integration was delayed by a bug appearing in a patch of a downstream dependency of swc (swc's tests would have caught it, but the release of swc happened before that bug). That dependency has been fixed and swc now pins to a specific version of that dependency.

dsherret:
Sad. This doesn't work in the following scenario:

```ts
const t = await test;
```

dsherret:
Completely implemented in 0.4.1 (published now)

PR #89, opened 2020-02-06
Support top level await

dsherret:
Implements #88.

Small issue in swc, so opened https://github.com/swc-project/swc/pull/630 and will wait for a new release.


dsherret:
Opened https://github.com/swc-project/swc/issues/632

Blocked by that for now.

Managed to get around the issue in PR 630 by ignoring empty statements, which is an improvement.

PR #95, opened 2020-02-10
Add preferHanging configuration for expressions

### LINKED ISSUE #73: Add preferHanging configuration for expressions
Right now by default, array literals will be hanging unless they are explicitly made multi-line. This should be changed to be multi-line by default and work similarly to the new `preferHangingArguments`. The code between these two areas should be shared as well because arrays are comma separated values.

dsherret:
Closes #73.

This will be a nice change.


dsherret:
Code is about 150ms slower on checker.ts. Overall, it's at 1140ms from 772ms a few weeks ago.

Not to worry though, I have some ideas to make this faster, but doing that is low priority at the moment... (ex. in the printer, look ahead conditions should only be jumped back to if necessary).

PR #162, opened 2020-03-26
basic Rust shell

bartlomieju:
This PR adds basic shell written in Rust to `dprint-plugin-typescript`.

CLI:
```shell
  dprint myfile1.ts myfile2.ts
  
  dprint --check myfile1.ts myfile2.ts
  
  dprint --config dprint.conf.json myfile1.ts myfile2.ts
```

<s>Reading config from JSON is still TODO</s>
I added deserialization from JSON to `dprint::configuration::Configuration` but go following error:
```
▶ target/debug/dprint_plugin_typescript --config ../../../deno/dprint.conf.json --check ../../../deno/std/testing/asserts.ts
missing field `useTabs` at line 4 column 1
```
I guess `Configuration` is fully resolved object... I need to think about it

### DIFF
--- a/packages/rust-dprint-plugin-typescript/Cargo.toml
+++ b/packages/rust-dprint-plugin-typescript/Cargo.toml
@@ -9,23 +9,13 @@ license = "MIT"
 homepage = "https://github.com/dsherret/dprint/tree/master/packages/rust-dprint-plugin-typescript"
 repository = "https://github.com/dsherret/dprint"
 
-[lib]
-name = "dprint_plugin_typescript"
-path = "src/lib.rs"
-
-[[bin]]
-name = "dprint_plugin_typescript"
-path = "src/main.rs"
-
 [dependencies]
-clap = "2.33.0"
 dprint-core = { path = "../rust-core", version = "0.12.0" }
 swc_common = "=0.5.9"
 swc_ecma_ast = "=0.18.1"
 swc_ecma_parser = "=0.21.6"
 swc_ecma_parser_macros = "=0.4.1"
 serde = { version = "1.0.88", features = ["derive"] }
-serde_json = "1.0"
 
 [dev-dependencies]
 dprint-development = { path = "../rust-development" }
--- a/packages/rust-dprint-plugin-typescript/Cargo.toml
+++ b/packages/rust-dprint-plugin-typescript/Cargo.toml
@@ -9,7 +9,16 @@ license = "MIT"
 homepage = "https://github.com/dsherret/dprint/tree/master/packages/rust-dprint-plugin-typescript"
 repository = "https://github.com/dsherret/dprint"
 
+[lib]
+name = "dprint_plugin_typescript"
+path = "src/lib.rs"
+
+[[bin]]
+name = "dprint_plugin_typescript"
+path = "src/main.rs"
+
 [dependencies]
+clap = "2.33.0"
 dprint-core = { path = "../rust-core", version = "0.12.0" }
 swc_common = "=0.5.9"
 swc_ecma_ast = "=0.18.1"
--- a/packages/rust-dprint-plugin-typescript/Cargo.toml
+++ b/packages/rust-dprint-plugin-typescript/Cargo.toml
@@ -25,6 +25,7 @@ swc_ecma_ast = "=0.18.1"
 swc_ecma_parser = "=0.21.6"
 swc_ecma_parser_macros = "=0.4.1"
 serde = { version = "1.0.88", features = ["derive"] }
+serde_json = "1.0"
 
 [dev-dependencies]
 dprint-development = { path = "../rust-development" }
--- a/packages/rust-dprint-plugin-typescript/src/main.rs
+++ b/packages/rust-dprint-plugin-typescript/src/main.rs
@@ -1,11 +1,76 @@
+extern crate dprint_core;
 extern crate dprint_plugin_typescript as dprint;
 
 use clap::{App, Arg};
 use dprint::configuration::{Configuration, ConfigurationBuilder};
 use serde_json;
+use std::collections::HashMap;
 use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
+
+fn main() {
+    let cli_parser = create_cli_parser();
+
+    let matches = cli_parser.get_matches();
+
+    let config = if let Some(config_path) = matches.value_of("config") {
+        let config_contents = match fs::read_to_string(&config_path) {
+            Ok(contents) => contents,
+            Err(e) => {
+                eprintln!("{}", e.to_string());
+                std::process::exit(1);
+            }
+        };
+
+        let unresolved_config: HashMap<String, String> =
+            match serde_json::from_str(&config_contents) {
+                Ok(c) => c,
+                Err(e) => {
+                    eprintln!("{}", e.to_string());
+                    std::process::exit(1);
+                }
+            };
+
+        let global_config_result =
+            dprint_core::configuration::resolve_global_config(&unresolved_config.clone());
+
+        if !global_config_result.diagnostics.is_empty() {
+            for diagnostic in &global_config_result.diagnostics {
+                eprintln!("{}", diagnostic.message);
+            }
+            std::process::exit(1);
+        }
+
+        let config_result =
+            dprint::configuration::resolve_config(&unresolved_config, &global_config_result.config);
+
+        if !config_result.diagnostics.is_empty() {
+            for diagnostic in &config_result.diagnostics {
+                eprintln!("{}", diagnostic.message);
+            }
+            std::process::exit(1);
+        }
+
+        config_result.config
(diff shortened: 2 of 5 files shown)


packages/rust-dprint-plugin-typescript/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    Ok(())
+}
+
+fn create_cli_parser<'a, 'b>() -> clap::App<'a, 'b> {
+    App::new("dprint")
+        .about("Format source files")
+        .long_about(
+            "Auto-format JavaScript/TypeScript source code.
+  
+  dprint myfile1.ts myfile2.ts
```
bartlomieju: Actually it's `dprint_plugin_typescript`

packages/rust-dprint-plugin-typescript/Cargo.toml
```diff
@@ -9,13 +9,23 @@ license = "MIT"
 homepage = "https://github.com/dsherret/dprint/tree/master/packages/rust-dprint-plugin-typescript"
 repository = "https://github.com/dsherret/dprint"
 
+[lib]
+name = "dprint_plugin_typescript"
+path = "src/lib.rs"
+
+[[bin]]
+name = "dprint_plugin_typescript"
```
dsherret: Instead of putting this in `rust-dprint-plugin-typescript` folder, could we do it in a new `rust-dprint` folder instead then have `dprint-plugin-typescript` as a dependency? I have the `dprint` crate and it would be good to make use of that.

packages/rust-dprint-plugin-typescript/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    let config = if let Some(config_path) = matches.value_of("config") {
+        let config_contents = match fs::read_to_string(&config_path) {
+            Ok(contents) => contents,
+            Err(e) => {
+                eprintln!("{}", e.to_string());
+                std::process::exit(1);
+            }
+        };
+
+        let c: Configuration = match serde_json::from_str(&config_contents) {
```
dsherret: Is there a way to deserialize the object to a list of key value pairs (`HashMap<String, String>`)? Then we could just get a configuration using `resolve_config` (https://docs.rs/dprint-plugin-typescript/0.9.5/dprint_plugin_typescript/configuration/fn.resolve_config.html) and output any diagnostics.

dsherret: Seems like it's very easy with serde actually: https://github.com/serde-rs/serde/issues/1387#issuecomment-420889991

packages/rust-dprint-plugin-typescript/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+}
+
+fn check_source_files(config: Configuration, paths: Vec<PathBuf>) -> Result<(), std::io::Error> {
+    let mut not_formatted_files = vec![];
+
+    for file_path in paths {
+        let file_path_str = file_path.to_string_lossy();
+        let file_contents = fs::read_to_string(&file_path).unwrap();
+        match dprint::format_text(&file_path_str, &file_contents, &config) {
```
dsherret: I really should just change this to just accept a `Path` object instead 🙂

I'll do that in the future.

packages/rust-dprint-plugin-typescript/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+    let f = if not_formatted_files.len() == 1 {
+        "file"
+    } else {
+        "files"
+    };
+    println!("Formatted {} {}", not_formatted_files.len(), f);
+    Ok(())
+}
+
+pub fn format(
```
dsherret: If this file is for the binary file, then should this be public?

packages/rust-dprint-plugin-typescript/src/main.rs
```diff
@@ -0,0 +1,189 @@
...
+        .arg(
+            Arg::with_name("files")
+                .help("List of file paths to format")
+                .takes_value(true)
+                .multiple(true)
+                .required(true),
+        )
+}
+
+fn main() {
```
dsherret: This is a huge nitpick (especially considering a lot of dprint is a mess at the moment 😅), but I'd like to make it anyway. When entering this file I was greeted with the `is_supported` function, but it would be nice if this function was at the top instead. That way it reads top down rather than bottom up. I know a lot of people are used to declare before use because of C, but I find it's kind of nice to read the high level code before the low level details 🙂

Anyway, not a big deal at all.

packages/rust-dprint-plugin-typescript/src/main.rs
dsherret: What's the command you guys are going to run in deno to format all the files without specifying them individually? Should we add an example of that here? (I'm not experienced on this topic)

packages/rust-dprint-plugin-typescript/src/main.rs
bartlomieju: It shouldn't

packages/rust-dprint-plugin-typescript/src/main.rs
bartlomieju: Yeah, no problem 👍

packages/rust-dprint-plugin-typescript/src/main.rs
bartlomieju: So `deno fmt` walks current directory recursively and then checks if they're matching supported extension in `is_supported` function. 

However for our internal code we're actually using `git ls-files` to get complete list of files to format and pass them to `prettier`.

packages/rust-dprint/src/main.rs
```diff
@@ -0,0 +1,213 @@
...
+            match serde_json::from_str(&config_contents) {
+                Ok(c) => c,
+                Err(e) => {
+                    eprintln!("{}", e.to_string());
+                    std::process::exit(1);
+                }
+            };
+
+        let global_config_result =
+            dprint_core::configuration::resolve_global_config(&unresolved_config.clone());
```
dsherret: Right now we don't have a global config, but in the future we could structure it like:

```jsonc
{
  // global config goes here (ex. "newLineKind": "auto")
  "typescript": {
    // typescript config goes here
  }
}
```

I think this should probably just pass in an empty hashmap in here for now.

dsherret: @bartlomieju I'm going to merge this change in later today after work. Other than this, it looks great! 😀

By the way, I only have a few issues remaining before I recommend using this on Deno's code: https://github.com/dsherret/dprint/milestone/2 -- Once those are done then it will be good to start using.

bartlomieju: @dsherret I'm not sure I understand... 
```json
{
    "lineWidth": "40"
}
```
^-- this config file works
```json
{
    "lineWidth": "40",
    "typescript": {}
}
```
^--- this config file doesn't, because `typescript` is a map not a string

I thought that different options can be mixed in a single file and running it through both global and TS config will pick up all options.

bartlomieju: > @bartlomieju I'm going to merge this change in later today after work. Other than this, it looks great! 😀
> 
> By the way, I only have a few issues remaining before I recommend using this on Deno's code: https://github.com/dsherret/dprint/milestone/2 -- Once those are done then it will be good to start using.

That's great! Waiting eagerly to replace prettier :))

dsherret: So there's "global config" that is config shared across all languages, then there's language specific config that's implemented in dprint-plugin-typescript.

For now we'll just have the typescript config at the top level like is here so it will be (as you've done here):

```ts
{
    "lineWidth": "40",
    "trailingCommas": "always"
}
```

In the future once we have a "global config" then we'll probably move the typescript config into a "typescript" property.

```ts
{
    "lineWidth": "40",
    "typescript": {
        "trailingCommas": "always"
    }
}
```

So since we don't have a global config at the moment in the json file, we should probably change this one line to `dprint_core::configuration::resolve_global_config(&new HashMap())`. Does that make sense?

dsherret: > I thought that different options can be mixed in a single file and running it through both global and TS config will pick up all options.

I think right now the `resolve_config` in `dprint-plugin-typescript` will pick up global config like `lineWidth` as well. The `resolve_config` only uses the global config for its defaults.

But yeah, now that I think about it, the code will work as-is right now so this is not a big deal.

bartlomieju: Okay, all clear, let me change that and link to this comment.

dsherret: Perfect! Thanks!

Yeah, here's a better example of what would work in the future:

```json
{
    "indentWidth": 4,
    "typescript": {
        "indentWidth": 2,
        "trailingCommas": "always"
    }
}
```

PR #330, opened 2021-01-02
chore: add dependabot to automate dependencies

chenrui333:
add dependabot support to automate the dependencies (as well as security dependency patches)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,7 +10,7 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     commit-message:
       prefix: "chore"
       include: "scope"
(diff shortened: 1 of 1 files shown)


dsherret:
Dependabot is my greatest enemy 😅. Causes so much noise.

@chenrui333 do you know if there is a way to limit this to critical security issues? This would be great if so!

chenrui333:
> Dependabot is my greatest enemy 😅. Causes so much noise.
> 
> @chenrui333 do you know if there is a way to limit this to critical security issues? This would be great if so!

we can make it monthly update or even limit the PRs. I am using it for OSS projects as well as my company projects, my experience on it is good so far. :)

dsherret:
Ok, let’s go for one month. That’s not too bad 🙂

Thanks!

dsherret:
Thanks @chenrui333!

dsherret:
Sorry, I reverted this right after it opened all the major version bumps PRs. I'd rather upgrade dependencies on my own time.

PR #345, opened 2021-03-14
Fix Windows platform path generation and matching issues

### LINKED ISSUE #344: `dprint v0.11.1` CLI file arguments and matching are broken for Windows
For the Windows platform, the `dprint` CLI is not expanding wildcards nor correctly matching files or globs.

```text
C:\...\dprint>dprint --version
dprint 0.11.1

C:\...\dprint>dprint --verbose check scripts\build-homepage.js
[VERBOSE]: Creating directory: C:\Users\Roy\AppData\Local\dprint\cache
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\cache-manifest.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugin-cache-manifest.json
[VERBOSE]: Checking path exists: ./.dprintrc.json
[VERBOSE]: Reading file: ./.dprintrc.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-typescript\dprint-plugin-typescript-0.40.3.cached
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-markdown\dprint-plugin-markdown-0.5.1.cached
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-json\dprint-plugin-json-0.8.0.cached
[VERBOSE]: Checking path exists: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-prettier\0.2.1\dprint-plugin-prettier.exe
[VERBOSE]: Globbing: ["scripts\\build-homepage.js", "!website/playground/build", "!scripts/build-website", "!**/node_modules", "!**/dist", "!**/target", "!**/*-lock.json"]
[VERBOSE]: Finished globbing in 10ms
No files found to format with the specified plugins. You may want to try using `dprint output-file-paths` to see which files it's finding.

C:\...\dprint>dprint --verbose check scripts\*.js
[VERBOSE]: Creating directory: C:\Users\Roy\AppData\Local\dprint\cache
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\cache-manifest.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugin-cache-manifest.json
[VERBOSE]: Checking path exists: ./.dprintrc.json
[VERBOSE]: Reading file: ./.dprintrc.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-typescript\dprint-plugin-typescript-0.40.3.cached
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-markdown\dprint-plugin-markdown-0.5.1.cached
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-json\dprint-plugin-json-0.8.0.cached
[VERBOSE]: Checking path exists: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-prettier\0.2.1\dprint-plugin-prettier.exe
[VERBOSE]: Globbing: ["scripts\\*.js", "!website/playground/build", "!scripts/build-website", "!**/node_modules", "!**/dist", "!**/target", "!**/*-lock.json"]
[VERBOSE]: Finished globbing in 10ms
No files found to format with the specified plugins. You may want to try using `dprint output-file-paths` to see which files it's finding.

C:\...\dprint>dprint --verbose check "scripts\*.js"
[VERBOSE]: Creating directory: C:\Users\Roy\AppData\Local\dprint\cache
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\cache-manifest.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugin-cache-manifest.json
[VERBOSE]: Checking path exists: ./.dprintrc.json
[VERBOSE]: Reading file: ./.dprintrc.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-typescript\dprint-plugin-typescript-0.40.3.cached
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-markdown\dprint-plugin-markdown-0.5.1.cached
[VERBOSE]: Checking path exists: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-prettier\0.2.1\dprint-plugin-prettier.exe
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-json\dprint-plugin-json-0.8.0.cached
[VERBOSE]: Globbing: ["scripts\\*.js", "!website/playground/build", "!scripts/build-website", "!**/node_modules", "!**/dist", "!**/target", "!**/*-lock.json"]
[VERBOSE]: Finished globbing in 13ms
No files found to format with the specified plugins. You may want to try using `dprint output-file-paths` to see which files it's finding.
```

rivy:
Fixes #344.

### DIFF
--- a/crates/dprint/Cargo.toml
+++ b/crates/dprint/Cargo.toml
@@ -28,6 +28,7 @@ twox-hash = "1.6.0"
 url = "2.1.1"
 wasmer = "=1.0.0"
 zip = "0.5.8"
+wild = "2.0.4"
 
 [target.'cfg(windows)'.dependencies]
 winreg = "0.7.0"
--- a/crates/dprint/src/environment/real_environment.rs
+++ b/crates/dprint/src/environment/real_environment.rs
@@ -99,8 +99,14 @@ impl Environment for RealEnvironment {
     fn glob(&self, base: &Path, file_patterns: &Vec<String>) -> Result<Vec<PathBuf>, ErrBox> {
         let start_instant = std::time::Instant::now();
         log_verbose!(self, "Globbing: {:?}", file_patterns);
+        let os_file_patterns = if cfg!(windows) {
+            file_patterns.iter().map(|v| v.replace("\\", "/")).collect::<Vec<String>>()
+        } else {
+            file_patterns.to_vec()
+        };
         let base = self.canonicalize(base)?;
-        let walker = globwalk::GlobWalkerBuilder::from_patterns(base, file_patterns)
+        let walker = globwalk::GlobWalkerBuilder::from_patterns(base, &os_file_patterns)
+            .case_insensitive(if cfg!(windows) { true } else { false })
             .follow_links(false)
             .file_type(globwalk::FileType::FILE)
             .build();
@@ -117,6 +123,7 @@ impl Environment for RealEnvironment {
             }
         }
 
+        log_verbose!(self, "File(s) matched: {:?}", file_paths);
         log_verbose!(self, "Finished globbing in {}ms", start_instant.elapsed().as_millis());
 
         Ok(file_paths)
--- a/crates/dprint/src/main.rs
+++ b/crates/dprint/src/main.rs
@@ -31,7 +31,7 @@ fn main() -> Result<(), ErrBox> {
 
 fn run() -> Result<(), ErrBox> {
     let stdin_reader = cli::RealStdInReader::new();
-    let args = cli::parse_args(std::env::args().collect(), &stdin_reader)?;
+    let args = cli::parse_args(wild::args().collect(), &stdin_reader)?;
     let environment = RealEnvironment::new(args.verbose, args.is_silent_output())?;
     let cache = Arc::new(cache::Cache::new(environment.clone())?);
     let plugin_cache = Arc::new(plugins::PluginCache::new(environment.clone()));


rivy:
I couldn't see any CLI tests, so I didn't add any tests corresponding to this fix.

crates/dprint/Cargo.toml
```diff
@@ -28,6 +28,7 @@ twox-hash = "1.6.0"
 url = "2.1.1"
 wasmer = "=1.0.0"
 zip = "0.5.8"
+wild = "2.0.4"
```
dsherret, line 31: An issue might be that it uses a different globbing dependency (https://crates.io/crates/glob) than what the rest of dprint uses (https://crates.io/crates/globwalk) and there seems to be some behaviour differences between the two. Perhaps this is not an issue though since it will cause the same behaviour on windows and unix for unquoted globbing.

Does this ignore quoted globs? From what I can tell from the source it seems to. If so, this looks good!

crates/dprint/src/environment/real_environment.rs
```diff
@@ -99,8 +99,14 @@ impl Environment for RealEnvironment {
     fn glob(&self, base: &Path, file_patterns: &Vec<String>) -> Result<Vec<PathBuf>, ErrBox> {
         let start_instant = std::time::Instant::now();
         log_verbose!(self, "Globbing: {:?}", file_patterns);
+        let os_file_patterns = if cfg!(windows) {
+            file_patterns.iter().map(|v| v.replace("\\", "/")).collect::<Vec<String>>()
+        } else {
+            file_patterns.to_vec()
+        };
```
dsherret, line 106: Actually, I was just thinking about this more and if someone uses backslashes in a config file then it won't work on linux. I'm going to revert this part of the change in order to force people to use forward slashses.

dsherret:
@rivy thanks for this fix!

crates/dprint/src/environment/real_environment.rs
rivy: I'll try to think about a cross-platform solution to that... I'll start by looking into what `git` does for *.gitignore*.

crates/dprint/Cargo.toml
rivy: 'wild' does ignore quoted globs.

Since 'wild' only affects the Windows command line args, I think the slight discontinuity between 'glob' and 'globwalk' is reasonable for the benefit that it brings. I don't think it will cause many user issues. That being said, I am hoping that Rust starts coming together with a more robust, common/std-supported glob/globwalk in the near future. There is some discussion about it (https://github.com/rust-lang-nursery/glob/issues/59), but it's slow going at the moment.

dsherret: @rivy cool. Yeah I did some testing after merging and it looks good! It's nice to have them acting the same from the command line now.

crates/dprint/src/environment/real_environment.rs
rivy: So, without this conversion, Windows paths are still broken.

```shell
C:\...\dprint>target\release\dprint.exe check crates\cli-core\src\*.rs --verbose
[VERBOSE]: Creating directory: C:\Users\Roy\AppData\Local\dprint\cache
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\cache-manifest.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugin-cache-manifest.json
[VERBOSE]: Checking path exists: ./.dprintrc.json
[VERBOSE]: Reading file: ./.dprintrc.json
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-markdown\dprint-plugin-markdown-0.5.1.cached
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-json\dprint-plugin-json-0.8.0.cached
[VERBOSE]: Checking path exists: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-prettier\0.2.1\dprint-plugin-prettier.exe
[VERBOSE]: Reading file: C:\Users\Roy\AppData\Local\dprint\cache\plugins\dprint-plugin-typescript\dprint-plugin-typescript-0.40.3.cached
[VERBOSE]: Globbing: ["crates\\cli-core\\src\\checksums.rs", "crates\\cli-core\\src\\lib.rs", "crates\\cli-core\\src\\terminal.rs", "crates\\cli-core\\src\\types.rs", "crates\\cli-core\\src\\url_utils.rs", "!website/playground/build", "!scripts/build-website", "!**/node_modules", "!**/dist", "!**/target", "!**/*-lock.json"]
[VERBOSE]: File(s) matched: []
[VERBOSE]: Finished globbing in 9ms
No files found to format with the specified plugins. You may want to try using `dprint output-file-paths` to see which files it's finding.
```

Maybe include it and note that the '/' *nix-style seperator is preferred for cross-platform use?

rivy:
I looked through the code for glob, globwalk, ripgrep, and wild. The general gist is converting the path separator on Windows hosts to the "/" *nix-style path separator.

How about something like e71fc73 as a fix?

It only changes command line arguments and leaves *nix-style path separators unchanged. As a bonus, it *might* work for other platforms with alternate path separators as well. And it leaves configuration arguments completely alone.

A notation in the documentation that configuration should use *nix-style globs as "cross-platform" would probably a good addition to this change.

dsherret:
@rivy I wonder about doing this on all platforms. So if someone does `"**\\*.ts"` in a configuration file or something similar via the command line, then it will work on linux as well and convert that to `**/*.ts`. Would that work or is there some edge case I'm not thinking of?

rivy:
So, after mulling this over for a while, part of the problem is that there is a conflation of file PATHS and glob PATTERNS.

Any automated modifications of glob PATTERNS will interfere with expressing some other pattern (such as matching files which include a backslash; eg, using `**\\*.ts`, which is wacky, but legal, for *nix platforms). Automatically converting `*\\*.ts` to `*/*.ts` changes the meaning from "files with backslashes ending in '.ts' in the current directory" to "files ending in '.ts' in any immediate subdirectory".

But, if you differentiate between PATHS and PATTERNS, much of the problem evaporates. When receiving a file PATH, no modifications need to be done (and globbing by `dprint` would not be needed); glob PATTERNS can also be left alone prior to expansion by `dprint`.

Arguments to `dprint` would be purely PATHS and then you could use `--includes` (and/or `--include`) and `--excludes` (and/or `--exclude`), both mirroring the configuration options, to add glob PATTERNS.

If you want to keep the current mixed PATH / PATTERN approach, we could test each argument as a simple file/directory match and avoid globbing when it exists (assume it's a PATH). This might be confusing for some edge cases, especially files which include glob characters, but would likely work for the large majority of arguments.

You can see a small modification of 932b343 to accomplish that comparison for the mixed PATH / PATTERN method  in 0fcff91.

--- 

On a related note, if I'm reading the `dprint` code correctly, I see a bug generating wrong files for command line supplied PATTERNs when using a configuration file located in a directory which is not the current directory. It looks like it would use the configuration file base directory as base for PATTERNs supplied, instead of the current directory. I expect that would be a surprise to a user.

dsherret:
@rivy Essentially at the moment, everything is a pattern... though the `[files]` description in the help is wrong.

> Arguments to dprint would be purely PATHS and then you could use --includes (and/or --include) and --excludes (and/or --exclude), both mirroring the configuration options, to add glob PATTERNS.

I think you're right about this. I like `--includes` matching the config file. Opened #348.

> Any automated modifications of glob PATTERNS will interfere with expressing some other pattern (such as matching files which include a backslash; eg, using **\\*.ts, which is wacky, but legal, for *nix platforms). Automatically converting *\\*.ts to */*.ts changes the meaning from "files with backslashes ending in '.ts' in the current directory" to "files ending in '.ts' in any immediate subdirectory".

Yeah this was the only scenario I could think of too and it's super rare. I doubt anyone will use backslashes in a source file's filename and then to want to do pattern matching looking at backslashes within the file name to determine whether dprint should format it or not. If this is the only scenario that could cause issues, I think we should not support this scenario and convert backslashes to forward slashes on all platforms. That way we get the same behaviour on all platforms instead of a glob working on Windows, but not on Linux or Mac.

So, thinking about it more, I'd rather it works on all platforms or doesn't work on all platforms, but in this case it seems ok to make it work on all platforms.

> On a related note, if I'm reading the dprint code correctly, I see a bug generating wrong files for command line supplied PATTERNs when using a configuration file located in a directory which is not the current directory. It looks like it would use the configuration file base directory as base for PATTERNs supplied, instead of the current directory. I expect that would be a surprise to a user.

There definitely is an issue here. I opened #347.

Thanks for your help and feedback!

rivy:
So, as a fix, you can:

#### Phase 1

1. revert the "wild::args" commit; plus 
2. change the documentation to specify that all command line arguments are gitignore-style glob PATTERNs (ie, `/` is always path separator, `\` is an escape).
3. (optional step) add `--includes PATTERN` / `--excludes PATTERN` options as transition to **phase 2**

As a happy continuing coincidence for *nix users, file arguments and shell glob expansion will generally match the basic glob PATTERN specification, so they get a little added convenience. Windows users are not as lucky and work a bit more, but at least have a well-specified and obtainable recipe.

Result -> fixed, compatible with current usage (the "v0.11 style"); (optionally) cross-platform when using `--includes` and `--excludes` options in place of arguments

#### Phase 2

In the future, a **phase 2** would be a breaking change, moving to true FILE arguments (which are *not* glob expanded by `dprint` [ie, use as is]) with `--includes PATTERN` and `--exclude PATTERN` options. PATTERNs *are* all glob expanded (relative to current directory for the command line options and relative to the config file for config patterns). The PATTERNs would continue to be in the gitignore-style (with `/` as the path separator and `\` as an escape) for *all* platforms. Add back in `wild::args`, giving shell expansion to Windows users, and you then have well specified cross-platform capabilities with added CLI convenience for *nix and Windows and minimal weird special cases or hacky conversion code.

Assuming a rust glob that works cross platform as specified with gitignore-style patterns, the **phase 2** code would, I think, "just work" across multiple platforms (even those with even weirder path separators) with no extra effort within `dprint`.

Result -> cross-platform, convenient CLI usage, minimal grey-decision-area code and ❤️🎉🚀

PR #363, opened 2021-05-29
chore: match plugin file fullname

TiceCosmos:


crates/dprint/src/cli/run_cli.rs
```diff
@@ -138,6 +138,16 @@ fn get_file_paths_by_plugin(plugins: &Vec<Box<dyn Plugin>>, file_paths: Vec<Path
...
+            }
+        }
+        if let Some(file_fullname) = file_path.file_name().and_then(|s| s.to_str()).map(|s| s.to_string()) {
+            if let Some(plugin) = plugins.iter().filter(|p| p.file_fullnames().contains(&file_fullname)).next() {
+                if let Some(file_paths) = file_paths_by_plugin.get_mut(plugin.name()) {
+                    file_paths.push(file_path);
+                } else {
+                    file_paths_by_plugin.insert(String::from(plugin.name()), vec![file_path]);
+                }
             }
```
dsherret, line 151: I think we should change this matching to prioritize plugins based on the order provided in the configuration file (so the order of `plugins` here). We could loop over file paths first, then plugins, then check the file extension and then the file path. This would allow people to specify a plugin like dprint-plugin-cargo-toml first and then a future plugin like dprint-plugin-toml... so it would format Cargo.toml with dprint-plugin-cargo-toml and then other .toml files with dprint-plugin-toml.

Alternatively, perhaps it should match based on the exact file name first for all plugins and then match based on the file extension.

Thoughts on what would be best?

crates/dprint/src/plugins/cache.rs
```diff
@@ -277,6 +277,7 @@ mod test {
             version: String::from("0.1.0"),
             config_key: String::from("test-plugin"),
             file_extensions: vec![String::from("txt"), String::from("dat")],
+            file_fullnames: vec![],
```
dsherret, line 280: We should add a test for this overall functionality.

dsherret (changes requested):
Thanks for this and it looks good! Just a few changes.

crates/dprint/src/cli/run_cli.rs
```diff
@@ -138,6 +138,16 @@ fn get_file_paths_by_plugin(plugins: &Vec<Box<dyn Plugin>>, file_paths: Vec<Path
                 } else {
                     file_paths_by_plugin.insert(String::from(plugin.name()), vec![file_path]);
                 }
+                continue;
+            }
+        }
+        if let Some(file_fullname) = file_path.file_name().and_then(|s| s.to_str()).map(|s| s.to_string()) {
+            if let Some(plugin) = plugins.iter().filter(|p| p.file_fullnames().contains(&file_fullname)).next() {
```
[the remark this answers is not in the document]
TiceCosmos: The meaning of different capitalization is the same.

crates/dprint/src/configuration/get_init_config_file_text.rs
```diff
@@ -79,13 +79,21 @@ pub fn get_init_config_file_text(environment: &impl Environment) -> Result<Strin
...
+            if !extension_includes.is_empty() {
+                json_text.push_str("\"**/*.{");
+                json_text.push_str(&extension_includes.join(","));
+                json_text.push_str("}\"");
+            }
+            if !file_name_includes.is_empty() {
+                json_text.push_str("\"**/{");
+                json_text.push_str(&file_name_includes.join(","));
+                json_text.push_str("}\"");
+            }
```
dsherret, line 96: Maybe add a test for when these are both set (extension and file_name)? I think there is a bug here unless I'm reading this wrong.

crates/dprint/src/cli/run_cli.rs
```diff
@@ -128,18 +128,30 @@ fn get_file_paths_by_plugin_and_err_if_empty(
...
-            }
-        }
+        let plugin = if let Some(plugin) = crate::utils::get_lowercase_exact_file_names(&file_path).and_then(|k| plugin_by_exact_file_name.get(k.as_str())) {
+            plugin
+        } else if let Some(plugin) = crate::utils::get_lowercase_file_extension(&file_path).and_then(|k| plugin_by_file_extension.get(k.as_str())) {
+            plugin
+        } else {
+            continue;
+        };
+        let file_paths = file_paths_by_plugin.entry(plugin.to_string()).or_insert(vec![]);
```
dsherret, line 153: Neat. I was wondering a little while ago what a short way of doing this is.

dsherret (changes requested):
Almost there! I think just one small issue and then I will merge this.

dsherret:
This looks good and sorry for my delay! I have been very busy the past couple weeks, but I will get to this on the weekend.

TiceCosmos:
> This looks good and sorry for my delay! I have been very busy the past couple weeks, but I will merge this on the weekend.

That’s all right. Thanks you.

dsherret (approved):

Issue #379, opened 2021-08-12
Globbing is slow for large monorepo


ryan-rushton:
Version: 0.15.1

We have a large monorepo and are trying to run some fast formatting commands on single files. Globbing is causing issues as it takes 4-6 seconds as part of the format for a single file.

```
dprint fmt --verbose --config ~/work/some-org/dprint.json ~/work/some-org/web/src/blah.json
[VERBOSE]: Creating directory: /Users/ryanrushton/Library/Caches/dprint/cache
[VERBOSE]: Reading file: /Users/ryanrushton/Library/Caches/dprint/cache/cache-manifest.json
[VERBOSE]: Reading file: /Users/ryanrushton/Library/Caches/dprint/cache/plugin-cache-manifest.json
[VERBOSE]: Reading file: /Users/ryanrushton/work/some-org/dprint.json
[VERBOSE]: Reading file: /Users/ryanrushton/Library/Caches/dprint/cache/plugins/dprint-plugin-json/dprint-plugin-json-0.12.3.cached
[VERBOSE]: Reading file: /Users/ryanrushton/Library/Caches/dprint/cache/plugins/dprint-plugin-markdown/dprint-plugin-markdown-0.9.4.cached
[VERBOSE]: Globbing: ["!**/node_modules", "!**/*-lock.json"]
[VERBOSE]: File(s) matched: []
[VERBOSE]: Finished globbing in 4331ms
[VERBOSE]: Writing file: /Users/ryanrushton/Library/Caches/dprint/cache/cache-manifest.json
[VERBOSE]: Reading file: /Users/ryanrushton/Library/Caches/dprint/cache/some-org.incremental
[VERBOSE]: Checking path exists: /Users/ryanrushton/Library/Caches/dprint/cache/some-org.incremental
[VERBOSE]: Creating instance of dprint-plugin-json
[VERBOSE]: Created instance of dprint-plugin-json in 1ms
[VERBOSE]: Reading file: /Users/ryanrushton/work/some-org/web/src/ui/home/search/search_bar/blah.json
[VERBOSE]: Formatted file: /Users/ryanrushton/work/some-org/web/src/ui/home/search/search_bar/blah.json in 0ms
[VERBOSE]: Writing file: /Users/ryanrushton/work/some-org/web/src/ui/home/search/search_bar/blah.json
Formatted 1 file.
[VERBOSE]: Writing file: /Users/ryanrushton/Library/Caches/dprint/cache/some-org.incremental
```

[...]

PR #380, opened 2021-08-13
editor service: keep resolved project files instead of evaluating each time

declanvong:
Resolves https://github.com/dprint/dprint/issues/379. 

Previously, the "check path" flow in the editor service would unconditionally re-glob the working directory. 
This PR makes it so that we glob the working directory once, on service startup, and cache the list of resolved files and file patterns from the config file.
"Check path" will now:
   - Reevaluate the config file to see if the includes / excludes has changed. If so, we re-resolve file paths. 
   - Otherwise, just check against the cached list. 
   - If we can't find the file, it may be a new file, so we re-resolve and try again (only once). 

This resolves the performance issue in the majority case, which is doing a quick format (e.g. on save) on an existing file, in a large codebase.

For reviewing, I recommend enabling GH's "Ignore whitespace" option for `get_and_resolve_file_paths` and below -- I've pulled them out from where they were originally inner functions, so it's pretty noisy from the indentation change. 

My first Rust project, please be gentle :)

### DIFF
--- a/crates/dprint/src/cli/run_cli.rs
+++ b/crates/dprint/src/cli/run_cli.rs
@@ -44,7 +44,7 @@ pub fn run_cli<TEnvironment: Environment>(
             plugin_pools.set_plugins(plugins);
             // if the path is absolute, then apply exclusion rules
             if environment.is_absolute_path(&cmd.file_path) {
-                let file_paths = resolve_file_paths(&config, &args, environment)?;
+                let file_paths = get_and_resolve_file_paths(&config, &args, environment)?;
                 // canonicalize the file path, then check if it's in the list of file paths.
                 match environment.canonicalize(&cmd.file_path) {
                     Ok(resolved_file_path) => {
@@ -67,7 +67,7 @@ pub fn run_cli<TEnvironment: Environment>(
         SubCommand::OutputFilePaths => {
             let config = resolve_config_from_args(&args, cache, environment)?;
             let plugins = resolve_plugins_and_err_if_empty(&config, environment, plugin_resolver)?;
-            let file_paths = resolve_file_paths(&config, &args, environment)?;
+            let file_paths = get_and_resolve_file_paths(&config, &args, environment)?;
             let file_paths_by_plugin = get_file_paths_by_plugin(&plugins, file_paths);
             output_file_paths(file_paths_by_plugin.values().flat_map(|x| x.iter()), environment);
             Ok(())
@@ -75,7 +75,7 @@ pub fn run_cli<TEnvironment: Environment>(
         SubCommand::OutputFormatTimes => {
             let config = resolve_config_from_args(&args, cache, environment)?;
             let plugins = resolve_plugins_and_err_if_empty(&config, environment, plugin_resolver)?;
-            let file_paths = resolve_file_paths(&config, &args, environment)?;
+            let file_paths = get_and_resolve_file_paths(&config, &args, environment)?;
             let file_paths_by_plugin = get_file_paths_by_plugin_and_err_if_empty(&plugins, file_paths)?;
             plugin_pools.set_plugins(plugins);
             output_format_times(file_paths_by_plugin, environment, plugin_pools)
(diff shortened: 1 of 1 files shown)


crates/dprint/src/cli/run_cli.rs
```diff
@@ -266,29 +266,69 @@ fn run_editor_service<TEnvironment: Environment>(
     let stdout = environment.stdout();
     let reader_writer = StdIoReaderWriter::new(stdin, stdout);
     let mut messenger = StdIoMessenger::new(reader_writer);
+    let config = resolve_config_from_args(&args, cache, environment)?;
+    let (mut file_patterns, mut absolute_paths) = get_config_file_paths(&config, args, environment)?;
+    let mut file_paths = resolve_file_paths(&file_patterns, &absolute_paths, &config, environment)?;
     let mut past_config: Option<ResolvedConfig> = None;
 
+    fn handle_check_path<TEnvironment: Environment>(
+        messenger: &mut StdIoMessenger<Box<dyn std::io::Read + Send>, Box<dyn std::io::Write + Send>>,
```
declanvong, line 275: Wasn't sure if there was a better way to pass these vars in -- I did try with a closure but handling the lifetimes because a little annoying.

crates/dprint/src/cli/run_cli.rs
```diff
@@ -266,29 +266,69 @@ fn run_editor_service<TEnvironment: Environment>(
...
+    ) ->  Result<(Vec<String>, Vec<PathBuf>, Vec<PathBuf>), ErrBox> {
+        let file_path = messenger.read_single_part_path_buf_message()?;
+        // check the config file paths to see if they have changed, re-resolve files in working directory if so
+        let config = resolve_config_from_args(args, cache, environment)?;
+        let (new_file_patterns, new_absolute_paths) = get_config_file_paths(&config, args, environment)?;
+        let new_file_paths = if !should_reattempt || &new_file_patterns != file_patterns || &new_absolute_paths != absolute_paths {
+            resolve_file_paths(&new_file_patterns, &new_absolute_paths, &config, environment)?
+        } else {
+            file_paths.clone()
+        };
```
declanvong, line 292: The primary change for this PR is here, and on lines 302-304.

crates/dprint/src/cli/run_cli.rs
```diff
@@ -665,90 +705,99 @@ fn resolve_plugins<TEnvironment: Environment>(
     return Ok(plugins);
 }
 
-fn resolve_file_paths(config: &ResolvedConfig, args: &CliArgs, environment: &impl Environment) -> Result<Vec<PathBuf>, ErrBox> {
+fn get_and_resolve_file_paths(config: &ResolvedConfig, args: &CliArgs, environment: &impl Environment) -> Result<Vec<PathBuf>, ErrBox> {
```
declanvong, line 708: Pulled this out into a new method, just so that we can return the file patterns / resolved paths and store them.

dsherret:
@declanvong this looks good and I will get to merging it soon!

I wonder though if we could just find out if a file matches based on the globs only (so we wouldn't have to consult the file system and get all the paths). We can look into that later as this is definitely a good improvement for now.

declanvong:
Yep definitely! That was raised on our end as well, and would also help out for the case where formatting from CLI, rather than from the editor service. It does look like it would be pretty simple with the standard `glob`: https://docs.rs/glob/0.2.11/glob/struct.Pattern.html#method.matches

I'll see if I can get a PR up for that soon as a separate improvement.

declanvong:
Hm, I've just taken a stab at the globset matching instead, and maybe it changes enough that it just supercedes this PR. Give me a little bit and I'll push it up to this branch for you to take another look.

declanvong:
Here's the draft: https://github.com/declanvong/dprint/pull/1/files

crates/dprint/src/cli/run_cli.rs
dsherret: This is ok! Yeah, sometimes when this happens it can help to either store all the state in a struct and then the functions can act on that struct.

For example, this could maybe all be extracted out to an `EditorService` struct that has a `start()` function on it or something. That would store all this state on the `EditorService` struct then `handle_check_path` could be a private implementation function on it. Anyway, don't worry about changing that in this PR as it's a bigger refactor that could maybe come down the line.

dsherret (approved):
This all looks good to me. Thanks @declanvong! I'm going to try it out and do a release right now.

dsherret:
@declanvong so unfortunately it doesn't seem to format a file if I do the following steps (with format on save):

1. Open an existing file, edit it, save.
2. Create a new file.
3. Save that file.

Instead of fixing this in this implementation, let's try to get https://github.com/declanvong/dprint/pull/1/files merged as I think it's a better solution and the code looks like it will work in there (or is very close to working). I appreciate the effort and work on this PR though 🙂

PR #382, opened 2021-08-13
chore: do not run external resource-dependent publishes in forks

joscha:
similar to https://github.com/dprint/dprint-plugin-typescript/pull/259


dsherret:
Thanks @joscha!

PR #386, opened 2021-08-20
editor service: check file against glob directly instead of walking

declanvong:


dsherret:
Looks great. Thanks again @declanvong! I'll do some additional testing then do a new release sometime this weekend.

Issue #396, opened 2021-08-27
Playground: allow loading custom wasm plugins


joscha:
Is there anything preventing us from allowing a custom plugin endpoint input to: https://dprint.dev/playground/?

E.g.:
![Playground_-_dprint](https://user-images.githubusercontent.com/188038/131141777-b1da440c-6e08-4245-8d7c-a8f3350ef348.png)

?

That could be really helpful in terms of developing plugins and/or forks.
If nothing prevents us I'd be happy to send a PR?

//cc @declanvong

joscha:
made a start here: https://github.com/dprint/dprint/pull/402

PR #402, opened 2021-08-30
feat: website: playground: allow custom wasm loading

joscha:
implements #396

Aug-31-2021 00-32-57

This is still a bit rough on the edges but wanted to get feedback first and also clarify a few things:

* can some of the data be read from the WASM file directly? Or can we add some generic data instead, to reduce the complexity of the input?
* is the approach with JSON and modal the way you envisioned it?
* I'd like to add loading from a file as well in additional?

### DIFF
--- a/website/playground/package-lock.json
+++ b/website/playground/package-lock.json
@@ -2025,6 +2025,11 @@
         }
       }
     },
+    "@sindresorhus/is": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-4.0.1.tgz",
+      "integrity": "sha512-Qm9hBEBu18wt1PO2flE7LPb30BHMQt1eQgbV76YntdNk73XZGpn3izvGTYxbGgzXKgbCjiia0uxTd3aTNQrY/g=="
+    },
     "@sinonjs/commons": {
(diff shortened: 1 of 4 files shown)


dsherret:
@joscha sorry, I haven't taken a look at the code in here yet. Will do so soon!

> can some of the data be read from the WASM file directly? Or can we add some generic data instead, to reduce the complexity of the input?

Basically all this information can be retreived from the wasm file now. I will merge a quick refactoring into this PR once I get a chance (going to assign it to me for a bit).

> is the approach with JSON and modal the way you envisioned it?

We'll be able to only provide a URL once i make the change above.

> I'd like to add loading from a file as well in additional?

Yes, that would be great!

dsherret:
@joscha I did an update to the playground in https://github.com/dprint/dprint/commit/1be03df9338d0f91ec3f1b5aa5b73d809d8e7ac1 so that it gets all the information it can based on the plugin url. This should make things a lot simpler for this PR.

I don't think I can help merge it into this branch though because last time I tried pushing to one of your branches on a PR I was not able to.

dsherret:
This is now out of date since all this information is retrieved from the wasm file now. I've just added a very quick MVP for this that closes #396.

Thanks for this contribution though, @joscha!

PR #405, opened 2021-08-31
fix: restore searching all directories when glob pattern does not contain a slash at the beginning or middle

### LINKED ISSUE #401: Globbing broken (`excludes`) in 0.16.1
Version: `0.16.1`

We have an excludes rule in the config in `0.15.0`:
```json
  "excludes": [
    "report.*.*.*.*.*.json"
  ]
```

that excludes some JSON that is created by IntelliJ:
```bash
.ijwb/report.20201030.234629.86995.0.001.json
.ijwb/report.20210827.130047.13633.0.001.json
.ijwb/report.20201030.235004.87316.0.001.json
.ijwb/report.20210827.132614.16001.0.001.json
# more here
````
Before, these files were ignored, with `0.16.1` they are now a match, being part of `output-file-paths` or `check`:

```
dprint output-file-paths --plugins https://plugins.dprint.dev/json-0.12.0.wasm https://plugins.dprint.dev/markdown-0.8.0.wasm
```

I thought it was maybe introduced in a825704706bb94a6b60a88dbd951f41f4d092552 and fixed again in bde47767604f71c6e129242191a6040226b5e730 but it seems to persist on `HEAD` (d78651ba4bf0cfc29e86d522da48b422186c3002). Will try and create a unit test for it.

### PR DESCRIPTION
**Pull request by dsherret**
This restores the behaviour back to the intended .gitignore behaviour.

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

Closes #401.

cc @joscha

### DIFF
--- a/crates/dprint/src/utils/glob_utils.rs
+++ b/crates/dprint/src/utils/glob_utils.rs
@@ -50,8 +50,8 @@ pub fn glob(environment: &impl Environment, base: impl AsRef<Path>, file_pattern
   Ok(results)
 }
 
-pub fn to_absolute_globs(file_patterns: &Vec<String>, base_dir: &str) -> Vec<String> {
-  file_patterns.iter().map(|p| to_absolute_glob(p, base_dir)).collect()
+pub fn to_absolute_globs(file_patterns: Vec<String>, base_dir: &str) -> Vec<String> {
+  file_patterns.into_iter().map(|p| to_absolute_glob(&p, base_dir)).collect()
 }
 
 pub fn to_absolute_glob(pattern: &str, dir: &str) -> String {
@@ -167,7 +167,7 @@ fn is_windows_absolute_pattern(pattern: &str) -> bool {
 
   // now check for the last slash
   let next_char = chars.next();
-  matches!(next_char, Some('/' | '\\'))
+  matches!(next_char, Some('/'))
 }
 
 pub struct GlobMatcherOptions {
@@ -225,8 +225,6 @@ mod tests {
     assert_eq!(is_absolute_pattern("!/test.ts"), true);
     assert_eq!(is_absolute_pattern("D:/test.ts"), true);
     assert_eq!(is_absolute_pattern("!D:/test.ts"), true);
-    assert_eq!(is_absolute_pattern("D:\\test.ts"), true);
-    assert_eq!(is_absolute_pattern("!D:\\test.ts"), true);
   }
 
   #[test]
--- a/crates/dprint/src/utils/glob_utils.rs
+++ b/crates/dprint/src/utils/glob_utils.rs
@@ -61,6 +61,12 @@ pub fn to_absolute_glob(pattern: &str, dir: &str) -> String {
   let mut pattern = pattern.replace("\\", "/");
   let dir = dir.replace("\\", "/");
 
+  // check to see if glob is negated
+  let is_negated = is_negated_glob(&pattern);
+  if is_negated {
+    pattern.drain(..1); // remove the leading "!"
+  }
+
   // .gitignore: "If there is a separator at the beginning or middle (or both) of
   // the pattern, then the pattern is relative to the directory level of the particular
   // .gitignore file itself. Otherwise the pattern may also match at any level below the
@@ -72,7 +78,7 @@ pub fn to_absolute_glob(pattern: &str, dir: &str) -> String {
 
   // trim starting ./ from glob patterns
   if pattern.starts_with("./") {
-    pattern = pattern.chars().skip(2).collect();
+    pattern.drain(..2);
   }
 
   // when the glob pattern is only a . use an empty string
@@ -82,12 +88,6 @@ pub fn to_absolute_glob(pattern: &str, dir: &str) -> String {
 
   // store last character before glob is modified
   let suffix = pattern.chars().last();
-
-  // check to see if glob is negated (and not a leading negated-extglob)
-  let is_negated = is_negated_glob(&pattern);
-  if is_negated {
-    pattern = pattern.chars().skip(1).collect(); // remove the leading "!"
-  }
 
   // make glob absolute
   if !is_absolute_pattern(&pattern) {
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: crates/dprint/src/utils/glob_utils.rs [hunk 1]
```diff
@@ -61,6 +61,15 @@ pub fn to_absolute_glob(pattern: &str, dir: &str) -> String {
   let mut pattern = pattern.replace("\\", "/");
   let dir = dir.replace("\\", "/");
 
+  // .gitignore: "If there is a separator at the beginning or middle (or both) of
+  // the pattern, then the pattern is relative to the directory level of the particular
+  // .gitignore file itself. Otherwise the pattern may also match at any level below the
+  // .gitignore level."
+  let is_relative = match pattern.find("/") {
```
joscha, line 68: negated globs follow this as well, no (this is a real question, I don't know the exact answer, my hunch is yes)? So a preceding `!` should be stripped if there is any?

crates/dprint/src/utils/glob_utils.rs
```diff
@@ -81,8 +90,12 @@ pub fn to_absolute_glob(pattern: &str, dir: &str) -> String {
   }
 
   // make glob absolute
-  if !is_absolute_path(&pattern) {
-    pattern = glob_join(dir, pattern);
+  if !is_absolute_pattern(&pattern) {
+    if is_relative || pattern.starts_with("**/") || pattern.trim().is_empty() {
+      pattern = glob_join(dir, pattern);
+    } else {
+      pattern = glob_join(dir, format!("**/{}", pattern));
```
joscha, line 97: this will break with negated globs

crates/dprint/src/utils/glob_utils.rs
```diff
@@ -153,7 +166,7 @@ fn is_windows_absolute_path(file_path: &str) -> bool {
 
   // now check for the last slash
   let next_char = chars.next();
-  next_char == Some('/')
+  matches!(next_char, Some('/' | '\\'))
```
joscha, line 169: is this tested against the real FS? If not, glob patterns are always posix?

crates/dprint/src/utils/glob_utils.rs
```diff
@@ -205,18 +218,28 @@ mod tests {
...
+
+    assert_eq!(to_absolute_glob("./test.ts", "/test/"), "/test/test.ts");
+    assert_eq!(to_absolute_glob("test.ts", "/test/"), "/test/**/test.ts");
+    assert_eq!(to_absolute_glob("*/test.ts", "/test/"), "/test/*/test.ts");
+    assert_eq!(to_absolute_glob("*test.ts", "/test/"), "/test/**/*test.ts");
+    assert_eq!(to_absolute_glob("**/test.ts", "/test/"), "/test/**/test.ts");
+    assert_eq!(to_absolute_glob("/test.ts", "/test/"), "/test.ts");
+    assert_eq!(to_absolute_glob("test/", "/test/"), "/test/**/test/");
+    // has a slash in the middle, so it's relative
+    assert_eq!(to_absolute_glob("test/test.ts", "/test/"), "/test/test/test.ts");
```
joscha, line 243: needs tests for negated globs

crates/dprint/src/utils/glob_utils.rs
joscha: ah, this is a CLI pattern that is passed as arg

joscha (commented):
This reminds me of what I had implmented before relying on `output-file-paths`. I might dig a second to see what edge cases I've encountered. I do remember a few more.

joscha:
> This restores the behaviour back to the intended .gitignore behaviour.

Just me as a Rust noob, ignore my crazy idea: there is not any chance we can load the original C-based code from git into dprint, right?

crates/dprint/src/utils/glob_utils.rs
dsherret: Thanks! I didn't think about negated globs at all.

uri-canva:
A good reference for this: https://github.com/BurntSushi/ripgrep/tree/master/crates/ignore, it's even published as its own crate: https://docs.rs/ignore/0.4.18/ignore/.

crates/dprint/src/utils/glob_utils.rs
dsherret: This should actually do the file slash conversion ahead of time and not bother with this. Will fix.

joscha:
If you want me to, I can sanity check this change against a 500+ LOC autogenerated dprint.json. I can diff `output-file-paths` of `0.15.0` against this and see what is different and why. I think I would feel more confident to get this right and backwards compatible if we were able to reuse an already-proven implementation and/or reuse the original impl from git, but that short regression test can possibly at least give us an indication.

dsherret:
> If you want me to, I can sanity check this change against a 500+ LOC autogenerated dprint.json. I can diff output-file-paths of 0.15.0 against this and see what is different and why. I think I would feel more confident to get this right and backwards compatible if we were able to reuse an already-proven implementation and/or reuse the original impl from git, but that short regression test can possibly at least give us an indication.

That would be great. Thanks! I can do a release after if everything is good.

> Just me as a Rust noob, ignore my crazy idea: there is not any chance we can load the original C-based code from git into dprint, right?

I think it would probably be possible. I'm not familiar with the original code though.

joscha:
> That would be great. Thanks! I can do a release after if everything is good.

Cool, let me know when you're done iterating and I'll try to run this - if you happen to have the binary at hand and pop it over via email (I sent you one) or other means of file sharing ;) that might actually be easiest

dsherret:
@joscha are you on linux, windows, or mac? (Will respond to your email tomorrow)

joscha:
> @joscha are you on linux, windows, or mac? (Will respond to your email tomorrow)

mac, but can run on linux as well if need be. Don't have a windows machine, soz :/

dsherret:
Here's a Mac build: [dprint.zip](https://github.com/dprint/dprint/files/7080540/dprint.zip)

joscha:
Preliminary results:

| dprint version | # matches |
|----------------|-----------|
| 0.15.1         | 30317     |
| 0.16.0         | 1         |
| 0.16.1         | 33336     |
| NEXT           | 30317     |


will compare now.

dsherret:
> The only difference noted is a performance one: 10s for 0.15.1 to do the globbing, 25s for NEXT.

Oof. Ok, will have to work on improving that. None of the repos I deal with are that big so globby hasn't been an issue. I'll open an issue after doing a release.

Thanks for verifying everything at least works ok!

Issue #416, opened 2021-09-21
Regression: globbing ignores trump later includes regardless of order


joscha:
Version: 0.17.1 but probably (untested) 0.16.0 and newer

Behaviour:
```json
{
  "$schema": "https://dprint.dev/schemas/v0.json",
  "json": {
  },
  "includes": [
    "!**/*.{json}",
    "**/a.json"
  ],
  "plugins": [
    "https://plugins.dprint.dev/json-0.13.0.wasm"
  ]
}
```

with a folder like:
```
-rw-r--r--   1 joscha  staff     0 21 Sep 12:47 a.json
-rw-r--r--   1 joscha  staff     0 21 Sep 12:47 b.json
-rw-r--r--   1 joscha  staff   198 21 Sep 12:52 dprint.json
```

yields an empty list with `output-file-paths`:

```
dprint --version
dprint 0.17.1

dprint output-file-paths
```

with `0.15.1` this correctly prints `a.json`:
```
❯ /Users/joscha/Downloads/dprint --version
dprint 0.15.1

❯ /Users/joscha/Downloads/dprint output-file-paths
/Users/joscha/work/dprint-glob-test/a.json
```

We use the original behaviour for opting out whole subtrees and then slowly opting, in, think of:

```glob
// format all JSON
**/*.json
// opt-out folder x
!x/**/*.json
// but format all in x/y
x/y/**/*.json
```

Please note that this regression has no workaround that I could find as of this moment. The glob order does not seem to matter whatsoever, ignores seem to always trump regardless of their position.

dsherret:
Interesting. I didn't know it behaved this way, but that makes sense to do. I'll mark it as a bug and do a patch release now.

joscha:
> I'll mark it as a bug and do a patch release now.

awesome! Thanks so much!

joscha:
I think the `0.17.2` behaviour is not 100% sane anymore now:

for:
```
-rw-r--r--   1 joscha  staff     0 21 Sep 12:47 a.json
-rw-r--r--   1 joscha  staff     0 21 Sep 12:47 b.json
-rw-r--r--   1 joscha  staff   215 22 Sep 12:11 dprint.json
```
with:
```jsonc
{
  "$schema": "https://dprint.dev/schemas/v0.json",
  "json": {
  },
  "includes": [
    // opt in all JSON
    "**/*.json",
    // opt-out all JSON
    "!**/*.json",
    // opt-in only a.json
    "a.json"
  ],
[...]

PR #419, opened 2021-09-22
Switch to using ignore crate

### LINKED ISSUE #417: Switch to using ignore crate for glob matching
As title.

dsherret:
Closes #417.

cc @joscha - You may wish to try out this PR. I'll do a release tomorrow.

On building, you may have to run:

```bash
cargo build --manifest-path=crates/test-plugin/Cargo.toml --release --target=wasm32-unknown-unknown
cargo build --manifest-path=crates/test-process-plugin/Cargo.toml --release

# this should produce the binary at ./target/release/dprint
cargo build --locked --all-features ---release
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -443,7 +443,7 @@ dependencies = [
  "dprint-cli-core",
  "dprint-core",
  "dunce",
- "globset",
+ "ignore",
  "jsonc-parser",
  "lazy_static",
  "num_cpus",
@@ -662,6 +662,24 @@ dependencies = [
  "matches",
  "unicode-bidi",
  "unicode-normalization",
+]
+
+[[package]]
+name = "ignore"
+version = "0.4.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "713f1b139373f96a2e0ce3ac931cd01ee973c3c5dd7c40c0c2efe96ad2b6751d"
+dependencies = [
+ "crossbeam-utils",
+ "globset",
+ "lazy_static",
+ "log",
+ "memchr",
+ "regex",
+ "same-file",
+ "thread_local",
+ "walkdir",
+ "winapi-util",
 ]
 
 [[package]]
@@ -1295,6 +1313,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "71d301d4193d031abdd79ff7e3dd721168a9572ef3fe51a1517aba235bd8f86e"
 
 [[package]]
+name = "same-file"
+version = "1.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "93fc1dc3aaa9bfed95e02e6eadabb4baf7e3078b0bd1b4d7b6b0b68378900502"
+dependencies = [
+ "winapi-util",
+]
+
+[[package]]
 name = "scopeguard"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1532,6 +1559,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "thread_local"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8018d24e04c95ac8790716a5987d0fec4f8b27249ffa0f7d33f1369bdfb88cbd"
+dependencies = [
+ "once_cell",
+]
+
+[[package]]
 name = "time"
 version = "0.1.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


dsherret:
> Let me know if you want me to test this against our extensive dprint.json to check for regressions :)

@joscha yes, that would be great! Thanks! I'm going to go to sleep now, but will check tomorrow.

I'll add you as a co-author on this when I merge.

joscha:
> @joscha yes, that would be great! Thanks! I'm going to go to sleep now, but will check tomorrow.

will do now. thanks for this!

joscha:
Two observations:

* This results in only `1` matched file, instead of `1926` with both 0.17.1 and 0.15.1
  This particular file is a file-specific inclusion after an exclusion of a subpath:
  ```glob
  !folder/v1/**/*
  folder/v1/README.md
  ```
  it is only one of a few though, so I am not sure why this is the only file that is matched, yet.
* The difference in glob times are substantial (that would make sense though, as 0.17.1 does not recurse into anything that was ignored at least once, and both pre-0.17.1 and c47952db56056684ade50f9d8a805b80d59c22d1 do): 
  `dprint output-file-paths` with
  * `0.15.1`: `3.56s user 5.34s system 60% cpu 14.696 total`
  * `0.17.1`: `4.58s user 4.36s system 109% cpu 8.181 total`
  * c47952db56056684ade50f9d8a805b80d59c22d1: `6.31s user 22.06s system 70% cpu 39.983 total`

joscha:
Actually, I can reproduce this with the test case mentioned in #416:

```
├── a
|  ├── a.json
|  └── b
|     └── b.json
└── dprint.json
```
with
```jsonc
{
  "$schema": "https://dprint.dev/schemas/v0.json",
  "json": {
  },
  "includes": [
    // opt in all JSON
    "**/*.json",
    // opt-out a
    "!a/**/*.json",
    // opt-in a/b
    "a/b/**/*.json"
  ],
  "plugins": [
    "https://plugins.dprint.dev/json-0.13.0.wasm"
  ]
}
```

yields nothing with c47952db56056684ade50f9d8a805b80d59c22d1 but 

```
a/b/b.json
dprint.json
```

with 0.15.1

declanvong:
Found the issue -- it's just a tiny mistake when using [`Override::matched`](https://docs.rs/ignore/0.4.18/ignore/overrides/struct.Override.html#method.matched) (or `OverrideBuilder::new` depending on how you look at it) :

>The given path is matched to the globs relative to the path given when building the override matcher. Specifically, before matching path, its prefix (as determined by a common suffix of the directory given) is stripped. If there is no common suffix/prefix overlap, then path is assumed to reside in the same directory as the root path for this set of overrides.


In `glob_matcher::build_override`, we're passing in `opts.base_dir` (e.g. `/some/path`) as the base. However, we also pass in the full absolute path for both the whitelist patterns, and the path to be matched during globbing. 

`ignore` strips the base from the path to be matched, so it ends up trying to compare the input path of `/a/a.json` against the patterns `/some/path/a/**/*.json` which obviously fails. 

This can be resolved by either simply using `OverrideBuilder::new("/");`, or making it so that the `includes` and `excludes` patterns are relative to the base directory instead.

I think the former makes more sense, i.e. just avoid `ignore`'s prefix stripping behaviour and work completely in absolute paths. That way we can do stuff like including patterns that are outside of the base directory completely, e.g. `/other/path`.

declanvong:
The test passes because it happens to use `"/"` as the working directory -- a simple shift to use a `/test/` subdirectory for everything causes it to fail:
image

dsherret:
Thanks @declanvong! Based on what I read in other repos, using all absolute paths is probably not ideal because it adds some overhead matching because there is a longer path to match against. I'll investigate if that's indeed the case later, but I suspect it is. For now I've done a bit of work and comitted it to the repo, but there is a bug in it. I ran out of time working on this tonight so I'll continue again tomorrow night. I think I just need to figure out what goes in the `build_override` function, but my brain is now mush so would be good to look at it a little fresher tomorrow.

dsherret:
I actually just had to come back to my computer for work and while I was here I think I fixed the issue. It would be worth trying this one out if anything to develop more test cases. I will look more into hopefully finishing this up tomorrow (later today).

joscha:
just tested - bd64dbb matches the same amount of files as 0.15.1 🎉

dsherret:
This is released now in 0.17.3. Thanks for the help once again @joscha and @declanvong!

I'm curious to know how long it takes to glob on your codebase with the latest changes? I will look at improving the performance.

joscha:
I think it's good:

```
/Users/joscha/Downloads/dprint-0.17.3 output-file-paths 1.55s user 2.01s system 136% cpu 2.610 total
```

started with a 10s one and then honed in on ~3s

PR #423, opened 2021-10-07
printer: implement rewrapping

declanvong:
Aims to resolve https://github.com/dprint/dprint-plugin-typescript/issues/52 and issues similar to it. Not production ready code, it's just a hack to see if it would work :) if the idea is sound, then i'll definitely be relying on you to implement it properly 😄 

See the test added here: https://github.com/dprint/dprint-plugin-typescript/pull/307/files#diff-43adf1891c4524ba7c0351e00d2e649ecc1211d4b14256d22aeb6f7c184193ceR1

### DIFF
--- a/crates/core/src/formatting/printer.rs
+++ b/crates/core/src/formatting/printer.rs
@@ -320,7 +320,7 @@ impl<'a> Printer<'a> {
         let rewrap_line_number = rewrap_save_point.writer_state.get_writer_info(self.writer.get_indent_width()).line_number;
         let save_point_info = save_point.writer_state.get_writer_info(self.writer.get_indent_width());
         let current_line_number = save_point_info.line_number;
-        let current_line_length = save_point_info.column_number - save_point_info.line_start_column_number;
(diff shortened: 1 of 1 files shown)


dsherret:
@declanvong sorry for my huge delay on this. I just rebased this (can't push to your branch because I don't have permission) and tried it out. It seems much better! I'm thinking we should add this here to the printer, but I want to try it out in larger codebases though.

declanvong:
sounds good! In that case, I can test it out in our repo first to do a double check. I'll also add you as a collaborator on my fork so that you can push straight up to the branch.

declanvong:
I'll also try to get some performance numbers on our repo -- I'm curious as to how much the additional save points + restores will affect total runtime. Hopefully it's a rare enough case that it doesn't impact anything majorly though.

dsherret:
Yeah, I failed at fixing the commit author 😅

declanvong:
Ran it over our codebase -- looking good still, after the rebase! Performance also looks no worse (in the few runs I tested, some were faster and some were slower, so it looks like any difference is negligible and down to run-by-run variance)

PR #428, opened 2021-10-20
Flip Into<PrintItems> impls to From

joshleeb:
This change flips the `impl Into<PrintItem> for _` to `impl From<_> for PrintItem`. The impact on the rest of the code is nil, and this is a non-breaking change to the public facing API.

The benefit is (as mentioned in the rustdocs for [`std::convert::From`](https://doc.rust-lang.org/std/convert/trait.From.html))

> One should always prefer implementing From over Into because implementing From automatically provides one with an implementation of Into thanks to the blanket implementation in the standard library.

There's an additional impl I added which says to implement `From<Option<T>>` for all `T`s where there is a `impl From<T> for PrintItems`. I thought that this would be better than `impl From<Option<PrintItemPath>> ...`, and any additional `From<Option<_>>`.

### DIFF
--- a/crates/core/src/formatting/print_items.rs
+++ b/crates/core/src/formatting/print_items.rs
@@ -7,6 +7,7 @@ use super::utils::{with_bump_allocator, CounterCell};
 
 /** Print Items */
 
+#[derive(Default)]
 pub struct PrintItems {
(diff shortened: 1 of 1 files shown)


joshleeb:
@dsherret and thanks for such a great Rust project 🎉

PR #432, opened 2021-11-04
Remove WriteItemsPrinterOptions

joshleeb:
This change refactors `WriteItemsPrinter` to remove `WriteItemsPrinterOptions`. I noticed that there were a few places performing a conversion between these types which are very similar. By removing `WriteItemsPrinterOptions` hopefully you'll agree that the `write_items` module, and upstream call sites are a little cleaner.


joshleeb:
Making this change also highlighted a potential optimization in the `WriteItemsPrinter`, to reduce the number of allocations when printing `WriteItems` and hopefully gain a small speed up.

More specifically, we know how many bytes each `WriteItem` variant is going to extend the output string by:

- for `NewLine`, `Tab`, and `Space` are trivially 1 byte;
- with `String` it is the size of the text in the `StringContainer`; and
- `Indent(n)` will be `n` for tabs, and `n * width` for spaces.

So, as we are adding items to the `WriterState` we can keep track of the expected, formatted output size and use that information to specify the strings capacity in `WriteItemsPrinter::print`. Something like (following this PR)

```rust
pub fn print<'a>(&self, items: WriteItems) -> String {
  items.fold(String::with_capacity(items.formatted_len), |acc, item| ...))
}
```

@dsherret wdyt? If this sounds good to you then I can put up a change and run some benchmarks to see if it makes a measurable difference (hopefully positive) 🤞

crates/core/src/formatting/writer.rs
```diff
@@ -1,9 +1,10 @@
...
 
-use super::collections::{GraphNode, GraphNodeIterator};
-use super::print_items::WriterInfo;
-use super::StringContainer;
-use super::WriteItem;
+use super::{
+  collections::{GraphNode, GraphNodeIterator},
+  print_items::WriterInfo,
+  StringContainer, WriteItem,
+};
```
dsherret, line 7: I've been gradually switching the code over to have one `use` statement per import because we do that in Deno to try to reduce the pain of merge conflicts. Not a big deal though. I will update this repo to use the unstable "imports_granularity" rustfmt option https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#Item%5C%3A

dsherret (approved):
Yup, this is cleaner. Thanks!

dsherret:
@joshleeb yeah that sounds like a good idea. We'd want to benchmark it to see if it's actually worth it though (I've found sometimes it's not in similar situations).

PR #433, opened 2021-11-05
Replace GraphNodeIterator with DoubleEndedIterator<Item=T>

joshleeb:
Please review after #432 has been merged.

With #427 being merged, deriving `Copy` on `WriteItem`, there is no need to iterate over `&GraphNode`. Instead, we replace `GraphNodeIterator` (which is an iterator over `&GraphNode`) with `GraphNode::iter` returning a `DoubleEndedIterator<Item=T>`.

This change also makes a few upstream call sites a bit nicer as well. For example

```diff
- items.into_iter().copied().collect::<Vec<_>>().into_iter().rev()
+ items.iter().rev()
```

in `print_with_allocator()`.


dsherret (approved):
Much nicer as per usual 🙂. I like the naming changes.

Thanks a lot @joshleeb! These code cleanups of dprint-core are great.

PR #452, opened 2022-01-26
Update README.md

IWANABETHATGUY:


dsherret:
Let's wait a bit before adding. I don't even have https://github.com/dprint/dprint-plugin-sql on there.

PR #456, opened 2022-01-28
Fix linter errors for tests in core

joshleeb:
This makes minor improvements by fixing clippy errors that occur when
linting the tests in crates/core.


dsherret (approved):
LGTM. Thanks @joshleeb! 😊

PR #460, opened 2022-02-02
core: replace CounterCell with IdCounter

joshleeb:
This change replaces `CounterCell` with a similar type `IdCounter`, and updates it's usages to better reflect the use case of generating unique, sequential ids.


dsherret (approved):
Looks good! Thanks @joshleeb!

PR #474, opened 2022-02-19
feat: exclude patterns are taken into consideration for all "include" CLI args

### LINKED ISSUE #473: dprint ignores excludes settings if files are passed directly (lint-staged)
we use dprint with lint staged:

```
//dprint.json
{
  "incremental": true,
  "lineWidth": 120,
  "indentWidth": 2,
  "useTabs": false,
  "typescript": {
    "quoteStyle": "preferSingle",
    "semiColons": "asi",
    "trailingCommas": "never"
  },
  "json": {},
  "markdown": {},
  "includes": ["**/*.{ts,tsx,js,jsx,cjs,mjs,json,md}"],
  "excludes": ["**/node_modules", "**/*-lock.json", "**/__generated__/**", "bazel*", "**/.next/**"],
  "plugins": [
    "https://plugins.dprint.dev/typescript-0.62.1.wasm",
    "https://plugins.dprint.dev/json-0.14.0.wasm",
    "https://plugins.dprint.dev/markdown-0.12.1.wasm"
  ]
}
```
```
//package.json
"lint-staged": {
    "**/*.{ts,tsx,js,jsx,cjs,mjs,json,md}": "dprint fmt"
  },
```

lint-staged will call dprint with staged filenames, but the problem is when those filenames are also part of the excluded files

ex:
`dprint fmt ./__generated__/some-generated-file.ts`

will format that file even though it should be excluded. this might be expected behavior, but it'd be nice to have a flag or setting to respect the excludes config even if files are passed by name

dsherret:
Previously, dprint would always format the provided file paths. This led to problems with tools like lint-staged where the excludes patterns weren't taken into account. See #473.

With this change, dprint treats all provided include and exclude CLI args as patterns instead of sometimes treating them as file paths (which had a lot of complexity and wasn't exactly right). The args simply override what is present in the config file (as occurred before, but non-glob file paths weren't ever excluded if they matched an excludes pattern).

Closes #473

FYI @joscha @declanvong (not sure this affects the setup you have… probably doesn’t?)

### DIFF
--- a/crates/core/src/formatting/print_items.rs
+++ b/crates/core/src/formatting/print_items.rs
@@ -2,9 +2,9 @@ use std::cell::UnsafeCell;
 use std::mem;
 use std::rc::Rc;
 
-use crate::formatting::id::IdCounter;
 use super::printer::Printer;
 use super::utils::with_bump_allocator;
+use crate::formatting::id::IdCounter;
 
 /** Print Items */
 
--- a/crates/dprint/src/commands/formatting.rs
+++ b/crates/dprint/src/commands/formatting.rs
@@ -904,10 +904,14 @@ mod test {
   #[test]
   fn should_error_when_no_files_match_glob() {
     let environment = TestEnvironmentBuilder::with_initialized_remote_wasm_plugin().build();
-    let error_message = run_test_cli(vec!["fmt", "**/*.txt"], &environment).err().unwrap();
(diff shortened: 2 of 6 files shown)


joscha:
nice one! Does this affect the service plugins for the editors?

dsherret:
@joscha no, that already should work this way (if the implementation of the client asks dprint if it can format the file first) so this would become consistent with that.

Ok cool. I'll do a release late tomorrow with this. I can always revert if we find this causes problems.

Issue #478, opened 2022-02-23
Support multi-threaded process plugins


dsherret:
This would help use multiple threads in a process plugin and allow for concurrent formatting instead of creating multiple processes.

For example, I want to make dprint-plugin-prettier use deno_core and when doing that have multiple isolates (or workers) formatting.

This would also be immediately useful for dprint-plugin-roslyn.

For now, this wouldn't be for Wasm plugins since Wasmer doesn't support threads with cranelift yet https://docs.wasmer.io/ecosystem/wasmer/wasmer-features

This should also support cancellation and be designed with #471 in mind.

PR #479, opened 2022-02-24
docs: documentation for process plugin schema version 4

dsherret:
This is the documentation for #478 that I will implement soon.

This schema version introduces cancellation, concurrent formatting requests, and range formatting. After this change is implemented, dprint will no longer create multiple processes for process plugins in order to do concurrent formatting. Instead, it will be up to the process plugin to do this. This will help reduce memory and should speed up formatting. The downside is that clients will be more complicated.

Todo:

- [x] Cancel host formats.


dsherret:
Closing in favour of documentation added in #486.

Issue #480, opened 2022-02-24
Add back tokio when implementing process plugin schema v4


dsherret:
Dprint currently has a custom worker stealing thread pool implementation that determines whether additional plugins should be created based on their startup time cost. It was fun implementing this, but there won't be any point to it soon. A few things have changed and are changing:

1. Wasmer is super fast at creating Wasm plugins now. Previously it took about 100ms on my old machine and then it took 5ms and on my current computer I think it takes 2ms or something.
2. Process plugins, once process plugin schema v4 are implemented, will only be created once (rather than multiple processes) and then they'll receive multiple formatting requests at the same time to parallelize.

So given that deciding whether to format something won't potentially have a huge performance cost (ex. creating many process plugins all at once), I think it makes sense to simplify this code a lot and just use tokio.

Issue #464, opened 2022-02-04
Support range formatting


azherebtsov:
Now that we have https://github.com/dprint/dprint-vscode and https://github.com/dprint/dprint-intellij and other editors support, it makes sense to have a way to format only portion of the file.
Possible use cases:
1. User selects a portion of the file in the IDE (editor) and triggers reformatting of this piece of text only
2. Reformatting only lines changed in a given commit
3. Reformatting a `*.diff` without even looking into original files (ranges of a diff can be reformatted by applying different plugins)

It might be beneficial for large code repositories where "re-format the world" approach is not an option.

Plugin to choose for a given range can be inferred from the file name as it is happening now.

dsherret:
It's a lot of work to implement this and get support in plugins. Additionally, most dprint plugins format files in under 10ms. For very large files like TypeScript's checker.ts (44,000 lines) I believe it took ~800ms last time I looked.

This change would also require a breaking change to the plugin interfaces and I'm not sure it's worth the complexity given how fast dprint is at formatting an entire file. That said, perhaps dprint editor plugins could attempt to format the selected range which should work in certain scenarios (ex. when formatting statements). The problem is that these plugins would need to have some context about what kind of file is being edited and it would probably just end up frustrating users.

azherebtsov:
Thank you David! 
These all are good arguments!
Another argument for range formatting is that (1) sometimes formatters are not consistent, and sometimes (2) the configuration of the formatter is changed. In both cases the engineer will try to avoid diffs in the lines they have not touched in a particular change (PR, commit), especially if they caused by an auto-formatter; therefore, they will want to run formatter on the limited range. 

In the case (2) it is not always desired to have "reformat the world commit", everybody hates them (lots of conflicts, hard to merge it, messed up history).

No pressure on this feature just now, as you correctly pointed execution time allows to reformat the whole file with the assumption that plugin output is consistent.

dsherret:
This has been fully implemented in dprint for a few months now as of (https://github.com/dprint/dprint/pull/858), but many of the plugins don't support it. Going to close this one as it's done in dprint itself.

PR #482, opened 2022-02-28
docs: editor-service schema version 5

dsherret:
This schema version introduces parallel messages, cancellable formatting, and range formatting support.

Before this is merged and worked on, the following tasks should be implemented:

- [x] Implement cancellation support and range formatting for process plugins (follow #479) - I have started working on this implementation and it is a big change so will take some time.
- [x] Switch to using tokio in the cli (#480)
- [x] ~~Look into and implement cancellation support and range formatting for Wasm plugins.~~ Will do this later

This won't be happening for a while, but FYI @ryan-rushton. Would appreciate your feedback in case you think there's anything missing or any features you think would be useful. I will help with the dprint-intellij changes when we get to that point.

Todo:

- [x] Align with the design in #486


ryan-rushton:
I will have a think about any other useful features but cancellation and range formatting are definitely at the top of my list for wanted features in the intellij plugin.

ryan-rushton:
@dsherret are we actually introducing range formatting? I was pointed to [this](https://github.com/dprint/dprint/issues/464#issuecomment-1030885956) conversation the other day.

dsherret:
Yes, I decided to add it after that feedback because I’m breaking all the interfaces right now in order to improve them. It’s going to have it in the platform at least, but not necessarily implemented by the existing plugins. I already implemented it in the new process plugin interface (along with cancellation). Will add it to the editor interface here next, then lastly to the Wasm plugin interface.

dsherret:
Implementation is in https://github.com/dprint/dprint/pull/505

PR #501, opened 2022-04-01
fix: mac aarch64

cliedeman:
fixed typo

Resolves https://github.com/dprint/dprint-plugin-roslyn/issues/8


cliedeman:
This doesn't completely fix things though but we do get further

```
Error resolving plugin https://plugins.dprint.dev/roslyn-0.6.3.exe-plugin: failed to fill whole buffer
```

dsherret (approved):
Oh wow! Thanks for tracking this down @cliedeman! There's no GitHub action M1 runners at the moment, but if they were available it would have caught this.

Curious about the second error. I will submit another commit to add more context to the errors.

dsherret:
This is released now in 0.24.3 along with improving the initial error reporting a bit.

PR #505, opened 2022-04-02
feat: implement editor schema v5

### LINKED ISSUE #471: Implement cancellation in the editor service
Probably would need a thread that always reads from stdin, which would allow the client to send a cancellation message and messages would need some identifier sent along with them that's included in the response.

dsherret:
Todo:

- [x] Implement this in the vscode editor https://github.com/dprint/dprint-vscode/pull/30
- [x] Implement this in the intellij plugin
- [x] Add a semaphore around formatting to limit the number of threads

Would be best to implement this with backwards compatibility in at least the vscode editor plugin.

Closes #471


crates/dprint/src/commands/editor/mod.rs
```diff
@@ -117,105 +132,128 @@ impl<'a, TEnvironment: Environment> EditorService<'a, TEnvironment> {
...
-        0 => return Ok(()),
-        // check path
-        1 => self.handle_check_path_message().await?,
-        // format
-        2 => self.handle_format_message().await?,
-        // unknown, exit
-        _ => bail!("Unknown message kind: {}", message_kind),
+      let message = EditorMessage::read(&mut self.reader)?;
+      // todo: temporary. Remove before merge or do a slimmed down version of it with log_verbose!(...)
+      eprintln!("Received message: {:?}", message);
```
dsherret, line 161: @ryan-rushton I added this message logging here and retested it with the vscode plugin and it's working.

PR #508, opened 2022-04-06
Fix installing from npm in paths with spaces

luisherranz:
The install script will fail if the current path contains spaces.

For example, the path `/Users/my name/dprint-project` will produce the error `/bin/sh: /Users/my: No such file or directory` when executing `/Users/my name/dprint-project/node_modules/dprint/install.sh x.x.x` with `exec`. The correct command should be `"/Users/my name/dprint-project/node_modules/dprint/install.sh" x.x.x`.


dsherret:
My bad! Thanks @luisherranz! Will do a release for this now.

luisherranz:
No problem. Thanks, David :)

PR #547, opened 2022-07-04
feat(cli): add --yes to dprint config update

midnightexigent:
this allows to disable prompts to confirm that we want to change url of the plugin

### DIFF
--- a/crates/dprint/src/arg_parser.rs
+++ b/crates/dprint/src/arg_parser.rs
@@ -77,7 +77,7 @@ pub struct FmtSubCommand {
 #[derive(Debug, PartialEq)]
 pub enum ConfigSubCommand {
   Init,
-  Update,
+  Update { yes: bool },
   Add(Option<String>),
 }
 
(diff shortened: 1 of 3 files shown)


crates/dprint/src/arg_parser.rs
```diff
@@ -380,6 +382,7 @@ EXAMPLES:
         .subcommand(
           Command::new("update")
             .about("Updates the plugins in the configuration file.")
+            .arg(Arg::new("yes").help("do not prompt to confirm to change url").short('y').long("yes").action(clap::ArgAction::SetTrue))
```
dsherret, line 385: We should change this message to be more clear about what they're agreeing to because there are security implications. Maybe "Upgrade process plugins without prompting to confirm checksums."

dsherret (commented):
Thanks @midnightexigent! This looks good to me, but would you be able to add a test for this in crates/dprint/src/commands/config.rs so we ensure this works and that it will continue to work in the future?

midnightexigent:
Hey. I didn't forget about this, I will try to do it this week

midnightexigent:
Should be good now

dsherret (approved):
LGTM. Thanks @midnightexigent!

PR #572, opened 2022-10-23
Make install_api.js work under Node 12

jakebailey:
I know that Node 12 is EOL, but I'd like to test it for some of my projects for now (plus for benchmarking reasons). But this piece of syntax exists in dprint's postinstall script, so if I install my deps, this will error out (even though I don't actually run dprint; it's just in my devDeps).

### DIFF
--- a/deployment/npm/install_api.js
+++ b/deployment/npm/install_api.js
@@ -198,7 +198,7 @@ function install() {
 module.exports = {
   runInstall() {
     return install().catch(err => {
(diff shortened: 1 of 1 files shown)


dsherret:
Thanks! Yeah, that's simple enough to support 😄

dsherret:
Published in 0.32.2

jakebailey:
Thanks!

PR #586, opened 2022-11-21
CI workflow: musl support

XDream8:
~~WIP~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -210,7 +210,7 @@ jobs:
     - name: Output checksums
       run: |
         echo "Linux x86_64 Zip: ${{needs.build.outputs.LINUX_X86_64_ZIP_CHECKSUM}}"
-        echo "Linux x86_64-musl Zip: ${{needs.build.outputs.LINUX_X86_64_ZIP_CHECKSUM}}"
+        echo "Linux x86_64-musl Zip: ${{needs.build.outputs.LINUX_X86_64_MUSL_ZIP_CHECKSUM}}"
         echo "Linux aarch64 Zip: ${{needs.build.outputs.LINUX_AARCH64_ZIP_CHECKSUM}}"
         echo "Mac x86_64 Zip: ${{needs.build.outputs.MAX_X86_64_ZIP_CHECKSUM}}"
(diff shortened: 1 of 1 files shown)


dsherret (commented):
@XDream8 I don't know much about musl. Are you able to run the current linux build on a musl system? I think we'll have to update the code more to support this (ex. search "x86_64" in the code) and then maybe we'll need to do some releases of process plugins to support this first.

XDream8:
> @XDream8 I don't know much about musl. Are you able to run the current linux build on a musl system?

nope current linux build does not work on a musl libc based system.
to be able to run it on a musl system you need to build it on a musl system or specify that target is a musl system like this;
```
rustup target add x86_64-unknown-linux-musl
cargo build -p dprint --locked --all-features --release --target x86_64-unknown-linux-musl
```

dsherret:
Ok thanks and good to know! We definitely should do this then.

We'll need to update the code to handle process plugins for this before we merge (btw, this pr for CI part looks great). I can work on it hopefully soon, but if you're able to contribute that would be helpful.

XDream8:
I have tested dprint on a musl libc based system a little these days and it seems that it works just fine since musl is just an implementation of C standart library.

Also since the x86_64 architecture is already supported dprint runs ok on a musl libc based system.

so this pr can be merged, this does not require additional support

dsherret:
I don't believe it will work with something like https://github.com/dprint/dprint-plugin-prettier ?

XDream8:
prettier plugin actually works with a musl libc based system
20221125_19h40m17s_grim

XDream8:
yep prettier plugin does not work just tested it

XDream8:
The easiest way to detect musl is checking if /lib/ld-musl-* does exist

I do not know of any way to do a wildcard file check with just std lib(I can just use a crate though). Do you know any way to do this?

dsherret:
> I do not know of any way to do a wildcard file check with just std lib(I can just use a crate though). Do you know any way to do this?

We can just iterate the directory entries and then check if any directory starts with that path. No need for another crate.

By the way, we'll also need to update https://github.com/dprint/dprint/tree/main/deployment to handle this.

XDream8:
I will see what i can do this weekend

XDream8:
Opened a new pr for all musl libc related code(#598)

PR #587, opened 2022-11-21
chore: upgrade clap to 4.0

dsherret:


crates/dprint/src/test_helpers.rs
```diff
@@ -147,29 +147,23 @@ USAGE:
     dprint <SUBCOMMAND> [OPTIONS] [--] [file patterns]...
 
 SUBCOMMANDS:
-    init                      Initializes a configuration file in the current directory.
-    fmt                       Formats the source files and writes the result to the file system.
-    check                     Checks for any files that haven't been formatted.
-    config                    Functionality related to the configuration file.
-    output-file-paths         Prints the resolved file paths for the plugins based on the args
-                                  and configuration.
```
dsherret, line 155: The wrapping in clap 4.0 isn't what I like because it doesn't seem to respect `.help_next_line(false)`, so I'm just not bothering with it.

Issue #591, opened 2022-11-28
Using system certificates for HTTPS connections


Pgolik:
Hello,
First of all, I am using this plugin, and really love simplicity and extensibility of it. Greate job!

ERROR:
While using dprint I have encountered an issue with certificates: 
`tls connection init failed: invalid peer certificate contents: invalid peer certificate: UnknownIssuer`.

FEATURE:
I would appreciate, if there would be a possibility to use system certs while downloading plugins.

ROOT CAUSE:
I am working in a project which is enclosed, and uses artifactory to store all required dependencies which currently in use. Artifactory got a set of certificates, which are added to system certs. Due to fact that those certificates are not in `webpki_roots::TLS_SERVER_ROOTS` provided by Mozilla, showed error is popping up. 

Thanks in advance 😄

tlewowski:
@dsherret would you be willing to accept a pull request for this? 

It seems like a simple change (adding `native-certs` feature to `ureq` features (`crates/dprint/Cargo.toml`), but if current behavior (fixed list of known certificates) is intended, it'd break.

dsherret:
I don't know enough about certificates to know what the proper behaviour should be here. Do most people use the system certs or mozilla ones? I feel like using system seems better. Maybe let's just do what you suggested and if it breaks anyone then we can make this configurable.

So yeah, a PR would be great. Thanks!

tlewowski:
most mature tools on Linux (git, curl etc.) either use system certs or have their own certificate configuration - so there is at least a possibility of running on a non-public setup (probably corporate networks are the most common case).
I'll issue a PR later today, thanks for the fast response!

dsherret:
Should be fixed in #599. Thanks @tlewowski! I'll do a release hopefully later today.

tlewowski:
@dsherret thank you! This will be useful for teams that are able to use one of the provided OS installers or `cargo`. For teams that use `npm`, we found one more problem with installation behind http proxy: https://github.com/dprint/dprint/issues/603

dsherret:
This is now configurable via the `DPRINT_CERT` and `DPRINT_TLS_CA_STORE` env vars: https://dprint.dev/setup/#tls-certificates

PR #599, opened 2022-12-05
#591 - use system-wide SSL certificates instead of fixed ones

tlewowski:
As discussed in #591 - use system SSL certificates instead of fixed set from Mozilla.

This should not significantly affect most users (Mozilla certificates should usually be available in trust root).


dsherret:
This is failing because the lockfile is out of date and I'm unable to push a fix because the change is being merged from the main branch:

```
To https://github.com/tlewowski/dprint.git
 ! [remote rejected]   tlewowski/main -> tlewowski/main (permission denied)
error: failed to push some refs to 'https://github.com/tlewowski/dprint.git
```

In the interest of time I'm going to merge this and then do another PR with the fix.

PR #617, opened 2023-01-11
fix(install): add support for the musl build in install script

wmnnd:
This change adds a check for musl to determine whether to download the glibc or musl version of dprint on x86_64 Linux.


dsherret (approved):
Thanks!

PR #626, opened 2023-01-30
fix(playground): don't use TS for 3rd plugins

g-plane:
Added `plaintext` for fallback when using 3rd plugins.


dsherret (approved):
LGTM. Thanks!

PR #628, opened 2023-01-31
feat(playground): save plugin to URL

g-plane:
I'm using this playground to debug a custom dprint plugin. However it doesn't save the plugin URL to browser URL, it will get lost when refreshing, which also makes hard to share a custom plugin to someone else.

This PR implements this feature.


website/playground/src/utils/UrlSharer.ts
```diff
@@ -65,15 +80,19 @@ export class UrlSaver {
...
     text: string;
     configText?: string;
+    plugin?: string;
     language?: "typescript" | "json" | "markdown" | "toml" | "dockerfile";
   }) {
     let url = `#code/${compressToEncodedURIComponent(text)}`;
     if (configText != null) {
       url += `/config/${compressToEncodedURIComponent(configText)}`;
     }
+    if (plugin != null) {
```
dsherret, line 93: Maybe we should avoid doing this if the plugin is a default plugin? Also, maybe we shouldn't append a language when the plugin is set?

dsherret (commented):
Thanks! I've wanted to do this, but never got around to it. I just have one comment.

website/playground/src/utils/UrlSharer.ts
g-plane: I've updated but added the logic outside of the URL saver.

dsherret (approved):
LGTM. Thanks @g-plane!

PR #638, opened 2023-02-28
feat(core): Implement From<GlobalConfiguration> for RecommendedGlobalConfiguration

stegaBOB:
Super quick PR to implement `From` on the `RecommendedGlobalConfiguration` struct.


dsherret (approved):
Thanks!

PR #645, opened 2023-03-11
refactor(core): group single line options into single struct

ekohilas:
For use by https://github.com/dprint/dprint-plugin-typescript/pull/502

> Currently, there is a multi line options struct that contains the options for multiple lines.
> 
> This PR aims to achieve a similar result by introducing a struct for single line options.


dsherret (approved):
Good idea. Thanks!

Issue #619, opened 2023-01-17
Install fails on aarch64 due to missing checksum


alexanderweiss:
Version: 0.34.1

In version 0.34, the checksum lookup for linux was updated to always include the linux family, but the checksum in `info.json` doesn't include the family for `aarch64`. It's looking for the key `linux-aarch64-gnu`, but the actual key is `linux-aarch64`. This causes installation to fail.

See https://github.com/dprint/dprint/commit/5a191b8ca1a2c1afc1f82914907a0ed188e1664f#diff-825a2a8dbcab67a42fe0f8bcd69f3b4bc199585cba51c745e16362f4e309191fL200-R146

dsherret:
Thanks for reporting this! Should be fixed in 0.34.4 and I modified the code to mitigate this in the future.

Also, sorry. I don't have a linux aarch64 machine and there are no github action runners for that yet in order to test the release

alexanderweiss:
Thanks for the quick fix! We have in the past used Docker with qemu to build ARM images on Github actions. That should work if you'd really like to have tests for it.

PR #650, opened 2023-03-12
fix: better cache

### LINKED ISSUE #618: Frequently experience dprint redownloading the plugins
Version: dprint 0.34.1

When I run `dprint fmt`, frequently it'll try to redownload the plugins even though I haven't changed anything afaik. This is particularly annoying for example if I lose internet connectivity or have bad internet connectivity because there's no fallback to use previously downloaded stuff.

Not sure if there's a log file I can attach to help debug or something.

### LINKED ISSUE #637: Save wasmer-compiler version with compiled wasm bytes
As title. It would help prevent issues like #636.

dsherret:
~~This saves a dprint magic header at the top of the serialized bytes that includes the wasmer-compiler version number in the format: `dprint<len(wasmer-compiler-version-number)><wasmer-compiler-version-number>`~~

Improves the cache code:

1. Incremental caches are stored at a hash of the root directory in the `incremental` folder
2. Remove files are stored at a hash of their url in the `remote` folder
3. The `cache-manifest.json` file is no longer needed. Removed.
4. Hashes of a file were happening twice with incremental. It only needs to happen once.
5. Serialized plugins are keyed on the underlying wasmer_compiler version to prevent issues like #619 
6. The cache is busted when upgrading for the first time, but not when downgrading (from this version on) in order to reduce the chance of #618.

Closes #637
Closes #618


dsherret:
Actually, the solution here should probably key on the version of wasmer-compiler to prevent #618. This will cause the cache size to grow over time, but #651 can help remedy that along with mentioned to run `dprint clear-cache` every now and again.

PR #661, opened 2023-04-10
Fix hang on install in Node 19+

jakebailey:
Node 19+ contain a bug where keepAlive'd sockets prevent the process from exiting in a timely manner without an explicit `process.exit()`. This leads to `npm install` hanging for 2 minutes running `install.js`. See: https://github.com/nodejs/node/issues/47228

For now, a workaround is to disable keepAlive. Another solution might be to add an explicit `process.exit(0)` somewhere, but I was less sure of that.

I don't think this is required for the proxy agent in the previous branch; as far as I'm aware, `https-proxy-agent` does not support `keepAlive`.

### DIFF
--- a/deployment/npm/install_api.js
+++ b/deployment/npm/install_api.js
@@ -91,6 +91,13 @@ function install() {
       const proxyUrl = getProxyUrl(url);
       if (proxyUrl != null) {
         options.agent = new HttpsProxyAgent(proxyUrl);
+      } else {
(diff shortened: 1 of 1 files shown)


dsherret (approved):
Interesting. Thanks! I’ll do a patch release after merging.

Edit: Actually, will do a release in the morning. Need to merge https://github.com/dprint/dprint/pull/662 first.

jakebailey:
Thanks, just tested and it works as it should.

PR #673, opened 2023-05-14
docs: update wasm plugin example

LeoDog896:
It:
- de-pins the serde patch version
- Uses the `sync` plugin handler


dsherret (approved):
Thanks!

PR #674, opened 2023-05-30
feat: add `allow-no-files` flag and config option

Ty3uK:
This option allows to exit `drpint` process with code `0` when there is some file passed to input and this file is also added to `excludes` property or cli arg.

For example, let's take `tsconfig.json`. I can mark it as excluded, but when my pre-commit hook will pass this file to `dprint fmt`, I got an error `No files found to format with the specified plugins`. But when I pass `--allow-no-files` or set this option in config, I can ran `dprint fmt` whout any issue.

### DIFF
--- a/crates/dprint/src/arg_parser.rs
+++ b/crates/dprint/src/arg_parser.rs
@@ -64,6 +64,7 @@ pub enum SubCommand {
 pub struct CheckSubCommand {
   pub patterns: FilePatternArgs,
   pub incremental: Option<bool>,
+  pub allow_no_files: bool,
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


disrupted:
should this even be a config option? I would imagine this could be default behavior with your example in mind.

Ty3uK:
> should this even be a config option? I would imagine this could be default behavior with your example in mind.

Well, I worried about breaking some existing code bases, so I thought about option with, maybe, some adoption time before making it default behaviour?

dsherret:
I think I'd rather it just returns a different exit code. Then you can ignore that exit code in the pre-commit hook.

Ty3uK:
> Thanks, but I think I'd rather it just returns a different exit code. Then you can ignore that exit code in the pre-commit hook.
> 
> I opened #675

Thanks!

PR #675, opened 2023-06-02
feat: fine grained exit codes

### LINKED ISSUE #384: Separate exit codes for real errors & assertion failure
Currently running `dprint check` with unformatted files exits `1`. I was trying to use this to distinguish code format assertions from other errors, like so:

```bash
local exit_code=0
dprint check || exit_code="$?"
if [[ "${exit_code}" -eq 1 ]]; then
  # we use a bespoke exit code when formatting fails
  # We chose a random one in the free space: https://tldp.org/LDP/abs/html/exitcodes.html
  exit_code=123
fi
exit "${exit_code}"
```

however, I just accidentally passed a non-existing configuration:

```bash
dprint check --config /path/to/nonexisting/config
```
which yields:

```
No files found to format with the specified plugins. You may want to try using `dprint output-file-paths` to see which files it's finding.
Error: Command failed: /Users/joscha/work/dprint check --config /path/to/nonexisting/config
```
and also has exit code `1`.

It would be great if the exit codes for errors and asserts were different. Maybe even different error codes for an actual panic, a config/setup error (like the one I made) and an assertion.

Thoughts?

dsherret:
Closes #674
Closes #384

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -544,6 +544,7 @@ dependencies = [
  "serde_json",
  "sha2",
  "similar",
+ "thiserror",
  "tokio",
  "tokio-util",
  "twox-hash",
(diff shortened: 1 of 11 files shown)


joscha:
Nice!

PR #678, opened 2023-06-27
feat: Add shell completions generator

zer0-x:
This should add support for generating [tab completion](https://en.wikipedia.org/wiki/Command-line_completion) scripts in run-time using `clap_complete`.

---

It could be done in compile-time, but first we need to isolate the `create_cli_parser` function in a separate file so we can `include!()` it in the `build.rs` file without any complexity.

### DIFF
--- a/crates/dprint/Cargo.toml
+++ b/crates/dprint/Cargo.toml
@@ -13,7 +13,7 @@ description = "Binary for dprint code formatter—a pluggable and configurable c
 [dependencies]
 anyhow = "1.0.69"
 clap = "4.3.8"
-clap_complete = { version = "4.3.1", optional = true }
+clap_complete = "4.3.1"
 console_static_text = "0.7.1"
 # don't upgrade to 0.26 https://github.com/crossterm-rs/crossterm/issues/752
 crossterm = "0.25.0" # manually retest everything when bumping this crate
@@ -51,7 +51,3 @@ winreg = "0.10.1"
 [dev-dependencies]
 path-clean = "0.1.0"
 pretty_assertions = "1.3.0"
-
-[features]
-default = ["completions_generator"]
-completions_generator = ["dep:clap_complete"]
--- a/crates/dprint/Cargo.toml
+++ b/crates/dprint/Cargo.toml
@@ -12,7 +12,8 @@ description = "Binary for dprint code formatter—a pluggable and configurable c
 
 [dependencies]
 anyhow = "1.0.69"
-clap = "4.1.4"
+clap = "4.3.8"
+clap_complete = { version = "4.3.1", optional = true }
 console_static_text = "0.7.1"
 # don't upgrade to 0.26 https://github.com/crossterm-rs/crossterm/issues/752
 crossterm = "0.25.0" # manually retest everything when bumping this crate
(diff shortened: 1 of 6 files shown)


crates/dprint/Cargo.toml
```diff
@@ -50,3 +51,7 @@ winreg = "0.10.1"
 [dev-dependencies]
 path-clean = "0.1.0"
 pretty_assertions = "1.3.0"
+
+[features]
+default = ["completions_generator"]
+completions_generator = ["dep:clap_complete"]
```
dsherret, line 57: Can we remove this feature and just make it required?

crates/dprint/src/arg_parser.rs
```diff
@@ -195,6 +197,19 @@ fn inner_parse_args<TStdInReader: StdInReader>(args: Vec<String>, std_in_reader:
     ("editor-service", matches) => SubCommand::EditorService(EditorServiceSubCommand {
       parent_pid: matches.get_one::<String>("parent-pid").and_then(|v| v.parse::<u32>().ok()).unwrap(),
     }),
+    #[cfg(feature = "completions_generator")]
+    ("generate-completions", matches) => {
```
dsherret, line 201: Let's just call this "completions" to align with what `deno` does (`deno completions`)

crates/dprint/src/main.rs
```diff
@@ -40,6 +40,12 @@ fn main() {
 
 async fn run(runtime_handle: tokio::runtime::Handle) -> Result<(), AppError> {
   let args = arg_parser::parse_args(std::env::args().collect(), RealStdInReader)?;
+
+  #[cfg(feature = "completions_generator")]
+  if let arg_parser::SubCommand::GenerateCompletions = args.sub_command {
+    std::process::exit(0);
```
dsherret, line 46:
```suggestion
    return Ok(());
```

Can we just do this?

crates/dprint/src/main.rs
```diff
@@ -40,6 +40,12 @@ fn main() {
...
+  }
```
dsherret, line 47: Can we remove this custom section and then just have the subcommand handler below handle this? Doing the extra work below of setting things up is not a big deal.

crates/dprint/src/arg_parser.rs
```diff
@@ -195,6 +197,19 @@ fn inner_parse_args<TStdInReader: StdInReader>(args: Vec<String>, std_in_reader:
...
+      let mut cmd = create_cli_parser(false);
+
+      clap_complete::generate(
+        matches.get_one::<clap_complete::Shell>("shell").unwrap().to_owned(),
+        &mut cmd,
+        "dprint",
+        &mut std::io::stdout(),
+      );
```
dsherret, line 209: Is it possible to have this be done below at?

https://github.com/dprint/dprint/pull/678/files#diff-50367ae714c483a503720af7c834d1c5297d348ee58b8a1878536cc007305874R136

Then we can reserve this function for just parsing and not functionality.

dsherret (commented):
Thanks! If you sign the CLA I can merge this soon.

crates/dprint/src/main.rs
zer0-x: It will not print any thing to `stdout` while setting things up, right?

dsherret: It shouldn't. If so I can fix any issue later.

dsherret (approved):
Looks great, thanks a lot @zer0-x!

(By the way, minor point, but I wanted to push some changes to this PR, but I can't because I can't push to your main branch. It makes things slightly easier to open PRs not from protected branches on other repos, but no big deal. I'll open up a separate PR to do a few slight improvements)

PR #689, opened 2023-07-03
fix: handle multiple processes writing to Wasm cache at the same time

dsherret:
Do an "atomic write" for more operations where it writes to a temporary file then renames.


dsherret:
Tested and this fixes the issue.

PR #694, opened 2023-07-04
fix: race condition where cache state got overwritten by other thread

dsherret:
This also makes the CI less flaky.


crates/dprint/src/commands/formatting.rs
```diff
@@ -287,7 +287,6 @@ mod test {
       .build();
     run_test_cli(vec!["fmt", "/file.txt"], &environment).unwrap();
     assert_eq!(environment.take_stdout_messages(), vec![get_singular_formatted_text()]);
-    assert_eq!(environment.take_stderr_messages().len(), 0);
```
dsherret, line 290: All of these are already implicitly checked for on the drop of the environment.

crates/dprint/src/plugins/cache.rs
```diff
@@ -191,7 +191,11 @@ impl<TEnvironment: Environment> ConcurrentPluginCacheManifest<TEnvironment> {
   }
 
   pub fn reload_from_disk(&self) {
-    *self.manifest.write() = read_manifest(&self.environment);
+    // ensure the lock is held while reading from the file system
+    // in order to prevent another thread writing to the file system
+    // at the same time
+    let mut manifest = self.manifest.write();
+    *manifest = read_manifest(&self.environment);
```
dsherret, line 198: Bug was here.

PR #698, opened 2023-07-06
feat: make using "includes" not the default

### LINKED ISSUE #695: Make the recommended workflow not require the "includes" pattern in dprint.json
The "includes" pattern makes things confusing.

For example, say someone has the prettier plugin and `"includes": "**/*.js"` in their dprint config. Then they run `dprint fmt "src/**/*.*"` on the command line... now it might be confusing that other files like json files are matched. In this case, someone really should be adding yaml and yml files to the excludes pattern so that it works the same way when running `dprint fmt` and `dprint fmt "src/**/*"`.

I believe the challenge here would be ensuring memory use doesn't get too high during globbing. We don't want to store a bunch of file paths in memory that won't be used as that could get expensive on very large repositories (though, maybe in those cases the recommended thing to do would be to use an "includes" pattern).

dsherret:
Closes #695

### DIFF
--- a/crates/dprint/src/paths.rs
+++ b/crates/dprint/src/paths.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::path::PathBuf;
 use std::str::Split;
 use thiserror::Error;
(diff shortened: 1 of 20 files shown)


crates/dprint/src/paths.rs
```diff
@@ -80,3 +94,27 @@ pub async fn get_and_resolve_file_paths(config: &ResolvedConfig, args: &FilePatt
...
+    result.push(format!("**/*.{{{}}}", file_exts.into_iter().map(|s| s.as_str()).collect::<Vec<_>>().join(",")));
+  }
+  if !file_names.is_empty() {
+    result.push(format!("**/{{{}}}", file_names.into_iter().map(|s| s.as_str()).collect::<Vec<_>>().join(",")));
+  }
+
+  // add the associations last as they are least likely to be matched
+  for plugin in plugins {
+    if let Some(associations) = plugin.get_config().0.associations.as_ref() {
+      result.extend(process_config_patterns(associations));
```
dsherret, line 116: This is wrong.

1. If there are positive association globs, then they take over completely from the file names and extensions.
2. If there are negative association globs, then they should be filtered out and not affect the file path globbing.

I should actually look into assigning paths to plugins at the same time as globbing to avoid mistakes and duplicated logic, though it might not be worth it.

Issue #702, opened 2023-07-09
Add Node.js plugins


dsherret:
I'd like to have the dprint-plugin-prettier plugin be executed with the system's Node.js binary in order to make it more portable.

```jsonc
// https://plugins.dprint.dev/prettier-0.27.0.json
{
  "schemaVersion": 3, // bumped from 2 -> 3
  "kind": "node",
  "name": "dprint-plugin-prettier",
  "version": "0.26.1",
  "archive": {
    "reference": "https://github.com/dprint/dprint-plugin-prettier/releases/download/0.27.0/dprint-plugin-prettier.zip",
    "checksum": "574f0d00106bdeabc4a2627cfce140d69ab387b75ae4a4a6d8275699916e9904"
  }
}
```

Two special named files in the zip:

1. `main.mjs` - This is the entrypoint and is required.
2. `snapshot.mjs` - If this file exists, it will be used for snapshotting to make starting up faster (this won't be implemented as part of the first pass).

This means that Node.js needs to be installed on the system, but it's pretty reasonable for certain plugins like the prettier plugin.

These would be categorized under "process plugins" and would use the same communication mechanism over stdio and use most of the same code.

dsherret:
It seems worker_threads in Node.js is slow beyond belief so I'm going to close this one for now and stick to using dprint_core in dprint-plugin-prettier.

unicornware:
@dsherret 

do you have any examples of dprint plugins written with [assemblyscript][]?

i would like to develop my own plugins, but i'm not a rust developer. dprint's ecosystem is not very welcoming to developers like myself.

[assemblyscript]: https://www.assemblyscript.org

PR #705, opened 2023-07-12
feat: node plugins

dsherret:
Requires way more tests before merging. It is working right now with the dprint-plugin-prettier rewrite.


dsherret:
https://github.com/dprint/dprint/issues/702#issuecomment-1633233013

Issue #711, opened 2023-07-15
Monorepos - Nested directory specific configuration


dsherret:
It would be useful to allow for directory specific configuration for large mono-repos.

```jsonc
// ~/.dprint.jsonc
{
  "typescript": {
    "indentWidth": 4,
    "newLines": "crlf"
  },
  "plugins": [
    "https://plugins.dprint.dev/typescript-0.85.1.wasm",
    "https://plugins.dprint.dev/json-0.17.4.wasm"
  ],
  "excludes": [
    // This project will not be formatted because the directory is not visited.
    "./project-never-visted/"
  ]
}

// ~/projects/project-a/dprint.json
{
  "inherit": true,
  "typescript": {
    // uses crlf newlines, and overrides to have indent 2
    "indentWidth": 2
  }
}

// ~/projects/project-b/dprint.json
{
  "inherit": true,
  "plugins": [
    // When the plugin url is different from the ancestor, it will NOT
    // inherit the the configuration for this plugin. This is because config
    // may differ between versions
    "https://plugins.dprint.dev/typescript-0.82.0.wasm"
  ]
}

// ~/projects/project-c/dprint.json
{
  "inherit": true,
  // plugins specified here have precedence over plugins in the root
  "plugins": [
    "https://plugins.dprint.dev/prettier-0.26.1.json@fdbe31f6aecd24f9d6b924214710a6766050d03146163b4e241e6056b2462f2e"
  ]
}

// ~/projects/project-d/dprint.json
{
  "inherit": true,
  // disable a plugin in a directory tree
  "typescript": {
    "associations": ["!**/*.*"]
  }
}

// ~/projects/project-e/dprint.json
{
  // this will not inherit plugins from the ancestor. In this case, only the toml files will be formatted
  "plugins": [
    "https://plugins.dprint.dev/toml-0.5.4.wasm"
  ],
}
```

1. Inheriting is opt-in, rather than opt-out. This is to prevent scenarios where a config file is added higher up and then suddenly starts affecting a sub config where that might not have been intentional.
2. Should nested configuration files inherit the includes or the excludes of the ancestor?
  * I'm not sure how someone would say "don't use the excludes of the ancestor"? Maybe `"inherit": { "excludes": false, "plugins": true }`....

Needs more thought...

It should be possible to do this very fast by collecting the config files while globbing.

dsherret:
This is implemented in 0.40, but doesn't have support for `"inherit": true` yet.

[...]

PR #719, opened 2023-07-24
feat: nested directory specific configuration

dsherret:
This does not implement inheritance of ancestor configs yet, but that should now be trivial to do.

Part of #711


dsherret:
There is no performance difference between this and main, however between main and the last release there has been a performance regression due to my refactorings in #712.

PR #740, opened 2023-08-20
docs: add ci caching docs

dsherret:


website/src/ci.md
```diff
@@ -8,10 +8,24 @@ layout: layouts/documentation.njk
...
+### Caching dprint's incremental cache
+
+You can get really fast formatting times on the CI by caching the `~/.cache/dprint` folder between runs on a Linux runner.
+
+```yml
+- uses: actions/cache@v3
+  with:
+    path: |
+      ~/.cache/dprint
+    key: ${{ runner.os }}-dprint-${{ hashFiles('**/dprint.json') }}
```
kachick, line 24: Do we need to specify the dprint version in this cache key?

In my experience, dprint `0.45.1` does not run the https://github.com/dprint/dprint/pull/858  compiled WASM plugin. But it works in `0.47.2`.

kachick: Ah, sorry for the noise. It will not be a cache problem.
In that case, I should bump the `dprint.json` to update the plugin versions. 🙇‍♂️

dsherret:
The dprint/check github action now natively supports caching. Going to close this.

PR #741, opened 2023-08-20
feat: add an lsp command

### LINKED ISSUE #738: Add `dprint lsp` command
Use https://crates.io/crates/tower-lsp -- It would be good to have this for support with more editors. Keep around the editor-service for now to evaluate the differences and then slowly migrate. I believe using the lsp will be faster, but have higher memory usage due to needing to keep the open files in dprint's memory.

Additional notes:

* Getting the text changes based on the input and output could be copied from here: https://github.com/denoland/deno/blob/b9f407aef73dd5773638a4ce1690737209318be8/cli/lsp/text.rs#L209
* Maybe [`MaskFutureAsSend`](https://github.com/dprint/dprint/blob/b56659b7c6a1095e58e61b0645d52723603f319b/crates/core/src/async_runtime.rs#L93) would be useful for just using refcell's everywhere, since tower-lsp requires Send + Sync, but dprint uses a single threaded runtime

MariaSolOs:
Closes #738 

_Don't get too excited yet. I'm just creating the PR early in case I need to refer to the code for questions. I'm still familiarizing myself with the code._

### DIFF
--- a/crates/dprint/src/commands/lsp/mod.rs
+++ b/crates/dprint/src/commands/lsp/mod.rs
@@ -1,6 +1,8 @@
 use std::rc::Rc;
 
 use dprint_core::plugins::process::start_parent_process_checker_task;
+use tokio::sync::oneshot;
+use tokio::try_join;
 use tower_lsp::jsonrpc::Result;
 use tower_lsp::lsp_types::DidChangeTextDocumentParams;
 use tower_lsp::lsp_types::DidCloseTextDocumentParams;
@@ -26,8 +28,14 @@ use crate::arg_parser::CliArgs;
 use crate::environment::Environment;
 use crate::plugins::PluginResolver;
 
+#[derive(Debug)]
+enum ChannelMessage {
+  Format(DocumentFormattingParams, oneshot::Sender<Result<Option<Vec<TextEdit>>>>),
+}
+
 struct Backend {
   client: Client,
+  sender: tokio::sync::mpsc::UnboundedSender<ChannelMessage>,
 }
 
 pub async fn run_language_server<TEnvironment: Environment>(
@@ -37,9 +45,32 @@ pub async fn run_language_server<TEnvironment: Environment>(
 ) -> anyhow::Result<()> {
   let stdin = tokio::io::stdin();
   let stdout = tokio::io::stdout();
+  let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
 
-  let (service, socket) = LspService::new(|client| Backend { client });
-  Server::new(stdin, stdout, socket).serve(service).await;
+  // tower_lsp requires Backend to implement Send and Sync, but
+  // we use a single threaded runtime. So spawn some tasks and
+  // communicate over a channel.
+  let recv_task = dprint_core::async_runtime::spawn(async move {
+    while let Some(message) = rx.recv().await {
+      match message {
+        ChannelMessage::Format(params, sender) => {
+          dprint_core::async_runtime::spawn(async move {
+            // todo: handle the params and format
+
+            // todo: return back an actual response
+            let _ = sender.send(Ok(None));
+          });
+        }
+      }
+    }
+  });
+
+  let lsp_task = dprint_core::async_runtime::spawn(async move {
+    let (service, socket) = LspService::new(|client| Backend { client, sender: tx });
+    Server::new(stdin, stdout, socket).serve(service).await;
+  });
+
+  try_join!(recv_task, lsp_task)?;
 
   Ok(())
 }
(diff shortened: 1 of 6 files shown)


crates/dprint/src/commands/lsp/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+use tower_lsp::lsp_types::TextEdit;
+use tower_lsp::LanguageServer;
+
+struct Backend;
+
+#[tower_lsp::async_trait]
+impl LanguageServer for Backend {
+  async fn initialize(&self, _: InitializeParams) -> Result<InitializeResult> {
+    Ok(InitializeResult {
+      // TODO: Any server info we want to include here?
```
dsherret, line 23: Awesome! Thanks so much for working on this.

I started very briefly setting this up so there's some code you can copy from here https://github.com/dprint/dprint/commit/3b3c4760b412061b802c87c3faeabf3690f1190f You might need to use a channel to get around tower lsp's send + sync requirements.

MariaSolOs: Thanks a lot! I very much welcome contributions and or helpful hints :)

> You might need to use a channel to get around tower lsp's send + sync requirements.

Not sure I understand what you mean :/ Isn't the LSP's connection enough?

dsherret: > Not sure I understand what you mean :/ Isn't the LSP's connection enough?

tower_lsp requires everything to implement the `Send` and `Sync` traits (so it works with a multi-threaded tokio runtime), but dprint uses a single threaded tokio runtime (I opened an issue earlier to drop this requirement in tower-lsp https://github.com/ebkalderon/tower-lsp/issues/386 which would make things more straightforward).

If you try to store any of dprint's structs on `Backend` it probably won't work because they're not `Send` or `Sync`. To get around that you could use a tokio channel (https://docs.rs/tokio/latest/tokio/sync/mpsc/fn.channel.html) to send messages from Backend to a task, process them, then send a response back to `Backend` on a oneshot channel. If you'd like, I can add a basic example to show how that would look.

MariaSolOs: Got it. Thanks a lot for explaining! I would indeed appreciate an example and I can try taking it up from there :)

dsherret: I pushed an example.

crates/dprint/src/commands/lsp/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  async fn did_close(&self, _: DidCloseTextDocumentParams) {
+    todo!()
+  }
+
+  async fn formatting(&self, params: DocumentFormattingParams) -> Result<Option<Vec<TextEdit>>> {
+    let (sender, receiver) = oneshot::channel();
+    let file_path = params.text_document.uri.to_file_path().expect("Expected a valid path");
+    let range = Default::default(); // TODO: Set it based on the file length
```
MariaSolOs, line 136: @dsherret just to make sure, when the range is `None` then it falls back to the entire document right?

crates/dprint/src/commands/lsp/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+    let file_text = "".into(); // TODO: Fetch from the latest document sync message
+    self.sender.send(ChannelMessage::Format(file_path, range, file_text, sender)).unwrap();
+    receiver.await.unwrap()
+  }
+
+  async fn range_formatting(&self, params: DocumentRangeFormattingParams) -> Result<Option<Vec<TextEdit>>> {
+    let (sender, receiver) = oneshot::channel();
+    let file_path = params.text_document.uri.to_file_path().expect("Expected a valid path");
+    let range = Default::default(); // TODO: Map LSP range to FormatRange
```
MariaSolOs, line 145: @dsherret what would you suggest in order to perform this mapping? Do we copy [the implementation from `deno`](https://github.com/denoland/deno/blob/b1ce2e41676ab5bc807a705b072986d9357fece5/cli/lsp/repl.rs#L259)?

dsherret: Deno internally uses this crate https://crates.io/crates/text_lines

I'm not sure it's worth computing and storing this though because it will do it for the whole document, which will probably be unnecessary. It can work for a first pass, but perhaps in the long term this could implement something custom.

MariaSolOs: Sounds good!

MariaSolOs:
After working on this for a while, I've realized that this will be more complex than I had first expected. I'm still studying the surrounding code and how to properly integrate the LSP into the existing architecture. I don't want to prevent someone else from working on this too, and so I'll close this PR for now and reopen it once I have a proper first implementation :)

crates/dprint/src/commands/lsp/mod.rs
dsherret: Yes, when the range is `None` then it formats the entire document.

dsherret:
@MariaSolOs yeah, this is complex. Would you be interested in implementing a subset of this feature like the document tracking or LSP ranges <-> byte index ranges?

MariaSolOs:
@dsherret of course :) I can work on the document tracking bits.

PR #747, opened 2023-09-08
fix: show anchor name in trace output

cjpearson:
Before:

Screenshot 2023-09-08 at 19 52 26

After:
Screenshot 2023-09-08 at 20 36 19


dsherret (approved):
Thanks!

PR #770, opened 2023-10-28
chore: try fix musl release build

### LINKED ISSUE #659: Support for prebuilt linux aarch64 musl
I'm attempting to install this this as an `npm` dependency on a node:alpine Docker image, but I get an error:

```
Error downloading dprint zip file. Error: Unknown status code 404 : Not Found
     at ClientRequest.<anonymous> (/app/node_modules/dprint/install_api.js:102:18)
```

I've given that stack trace, it looks like the url it's looking to download the dprint zip file looks something like this:

```
https://github.com/dprint/dprint/releases/download/dprint-aarch64-unknown-linux-musl.zip
```

To reproduce, you can build a docker image with this Dockerfile:

```Dockerfile
FROM node:18-alpine
RUN mkdir -p /app
WORKDIR /app
RUN echo '{"devDependencies":{"dprint":"*"}}' > package.json
RUN yarn install
```

Then run:

```sh
docker build --platform linux/arm64 .
```

I'm running this on an M1 mac, so I'm unsure if running this will work on non-arm devices.

I'm also unsure if this is as easy as just adding it to the list of platforms here https://github.com/dprint/dprint/blob/main/.github/workflows/ci.generate.ts and regenerating that that ci.yaml workflow.

dsherret:
Closes #659

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,8 +1,3 @@
 # statically link the c runtime https://github.com/rust-lang/rust/issues/100874
 [target.'cfg(all(windows, target_env = "msvc"))']
 rustflags = ["-C", "target-feature=+crt-static"]
-
-# https://github.com/rust-lang/rust/issues/46651#issuecomment-433611633
-[target.aarch64-unknown-linux-musl]
-linker = "aarch64-linux-gnu-gcc"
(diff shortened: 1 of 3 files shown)


dsherret:
Amazing. https://github.com/cross-rs/cross is great.

Issue #790, opened 2023-11-26
Do not convert file text to string in CLI


dsherret:
There's really no need. Keep it as `Vec<u8>` and send it to the plugins.

https://github.com/dprint/dprint/blob/e868948411f2e10a4390e32c6baaf8f779b89c91/crates/dprint/src/plugins/plugin.rs#L38

PR #792, opened 2023-11-26
feat: support formatting `Vec<u8>` (not just utf-8 strings)

dsherret:
This allows dprint (as a platform) to format any kind of data by storing the files as `Vec<u8>` instead of utf-8 strings. Currently most of the plugins still assume utf-8 strings.

Probably also a slight non-noticable perf improvement because the CLI no longer needlessly converts from `Vec<u8>` to `String` except when outputting diffs.

Closes https://github.com/dprint/dprint/issues/790


dsherret:
> Probably also a slight non-noticable perf improvement

Yes, on dprint's repo `dprint fmt` went from 117ms to 111ms.

Issue #801, opened 2023-12-14
Regression - 0.43.2 finds fewer files than 0.43.1


bradzacher:
It looks like #797 was a bit further-reaching than intended.
When I run `dprint output-file-paths | wc -l` on Canva's codebase:
- 0.43.1 - 83,896 files
- 0.43.2 - 69,695 files

Cases like this aren't being handled any more:
```json
{
  "include": [
    "!src/foo/**/*",
    "src/foo/file.ts"
  ]
}
```

Our config setup is quite ugly and complicated, sadly.
We have an auto-generated config that allows teams to configure formatting using `.gitattributes` files. In some cases people will opt out a folder higher-up, and then opt-in specific files or specific sub-folders.

As an example one might opt-out all files in `src/foo` because it contains generated files, but specifically opt-in `src/foo/README.md` because that's a human-written file.

dsherret:
Ah, I remember that now and thanks for spotting this! I'll add a test for this scenario and look into this soon.

dsherret:
@bradzacher I was looking at this more and realized the CLI args just overridding what's in the config isn't a good design. I opened https://github.com/dprint/dprint/pull/804 -- Does what's done there make sense? Let me know if you have any suggestions.

bradzacher:
That seems logical to me.

Based on what I've seen internally - all the usecases we have for CLI args is exactly as you describe - we are instructing dprint to format a subset of the allowed files and nothing more.

I don't even think people use the exclude arg - they just run `dprint fmt "some/folder/**/*"` to ensure their changes are good before submitting them.

dsherret:
Great. I'm going to merge this and do a minor release tonight. I'm pretty confident that most people would expect this behaviour.

> Also worth noting that from CI we also don't even pass globs any more or call output-file-paths - we have got it fast enough to just run dprint fmt on CI!!

That's great to hear and it's the way it should be! Way less complexity.

dsherret:
Should be fixed now in the latest release.

PR #804, opened 2023-12-18
feat: change CLI file patterns to be intersection with the config and excludes to be a union

dsherret:
The current behaviour of how the CLI file patterns just override the config is very confusing and doesn't work well.

This change does the following:

1. Paths provided to the CLI are now an intersection with the config includes patterns.
   - In other words, this only matches file patterns provided as CLI args that also match the config.
2. Paths provided as `--excludes` are now a union with the config excludes patterns.
   - In other words, this excludes patterns in addition to what's excluded in the config.
3. Added a new `--includes-override` flag for overriding the value in the config. This is the previous behaviour of providing includes on the CLI.
4. Added a new `--excludes-override` flag for overriding the "excludes" value in the config. This is the previous behaviour of `--excludes`.

Closes https://github.com/dprint/dprint/issues/801

### DIFF
--- a/crates/dprint/src/arg_parser.rs
+++ b/crates/dprint/src/arg_parser.rs
@@ -153,8 +153,10 @@ pub enum HiddenSubCommand {
 
 #[derive(Debug, Default, PartialEq, Eq)]
 pub struct FilePatternArgs {
-  pub file_patterns: Vec<String>,
-  pub exclude_file_patterns: Vec<String>,
+  pub include_patterns: Vec<String>,
+  pub include_pattern_overrides: Option<Vec<String>>,
+  pub exclude_patterns: Vec<String>,
+  pub exclude_pattern_overrides: Option<Vec<String>>,
   pub allow_node_modules: bool,
 }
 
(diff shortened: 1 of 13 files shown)


crates/dprint/src/arg_parser.rs
```diff
@@ -580,13 +588,27 @@ impl ClapExtensions for clap::Command {
...
       .arg(
         Arg::new("excludes")
           .long("excludes")
           .value_name("patterns")
+          .help("List of file patterns or directories in quotes to exclude when formatting. This excludes in addition to what is found in the config file.")
+          .num_args(1..),
+      )
+      .arg(
+        Arg::new("excludes-override")
+          .long("excludes-override")
```
dsherret, line 610: Not entirely sure about this name. Another alternative: `--config-excludes` (but the problem that has is it's ambiguous where the base of the pattern is--relative to the cwd or config file?)

I think overall people won't be using this flag that much and I mostly kept it to maintain functionality with the previous versions, so maybe not a big deal whatever it's called.

PR #820, opened 2024-02-11
Allow building from source if no prebuilt binary is available

hack3ric:
Prebuilts are limited to several platforms, so it is better to support building dprint from source when they are not available on current platform. As TypeScript now requires dprint to build, it should ease new platforms like RISC-V to build it.

WebAssembly plugins are portable across platforms. Native plugins (like `dprint-plugin-prettier` and `-exec`) still rely on fixed prebuilts. Is there any elegant solution to it? I would like to help and contribute my code for extending dprint's cross-platform ability, if there's a good way of doing it.

### DIFF
--- a/deployment/npm/install_api.js
+++ b/deployment/npm/install_api.js
@@ -4,11 +4,13 @@
 const fs = require("fs");
 const os = require("os");
 const path = require("path");
+const https = require("https");
+const cp = require("child_process");
 /** @type {string | undefined} */
 let cachedIsMusl = undefined;
 
 module.exports = {
-  runInstall() {
+  async runInstall() {
     const dprintFileName = os.platform() === "win32" ? "dprint.exe" : "dprint";
     const targetExecutablePath = path.join(
       __dirname,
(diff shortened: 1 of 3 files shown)


deployment/npm/install_api.js
```diff
@@ -20,8 +22,24 @@ module.exports = {
...
+      console.log("Building dprint binary from source...");
+      const packageVersion = require("./package.json").version;
+      const tmp = fs.mkdtempSync(path.join(os.tmpdir(), "tmp-"));
+      const dprintZip = path.join(tmp, "dprint.zip");
+      const dprintRepo = path.join(tmp, `dprint-${packageVersion}`);
+      await downloadFile(
+        `https://github.com/dprint/dprint/archive/refs/tags/${packageVersion}.zip`,
+        dprintZip,
+      );
+      await unzipFile(dprintZip, tmp);
```
dsherret, line 36: Maybe it should run `cargo install --locked --version VERSION_GOES_HERE dprint` instead and only if `DPRINT_BUILD_FROM_SOURCE` is set? That way it doesn't need a dependency on yauzl.

Also, what platform are you running on that dprint doesn't have a build for? Maybe we can try to add a distribution for that instead?

hack3ric: I am currently working on Arch Linux RISC-V and I would appreciate if you consider adding linux-riscv64-{gnu,musl} distributions for dprint and its native plugins. Besides, running WASM plugins like TypeScript formatter on riscv64 could produce inconsistent results when I tried to build tsc, which definitely needs further investigation...

As for the change, I lost access to my main work PC recently and will get back to it a few days later. :)

dsherret:
riscv64-unknown-linux-gnu support is now available in dprint 0.47.6. Let's instead try to add binaries to dprint's release instead of this approach.

PR #823, opened 2024-02-18
docs: update detection of AST specific config properties

zOadT:
The generated default config in the [playground](https://dprint.dev/playground/#code/Q/plugin/typescript) is slightly inconsistent:
* Some properties are missing:
  * `module.sortImportDeclarations`
  * `module.sortExportDeclarations`
  * `exportDeclaration.sortNamedExports`
  * `importDeclaration.sortNamedImports`
  * `exportDeclaration.forceSingleLine`
  * `importDeclaration.forceSingleLine`
  * `exportDeclaration.forceMultiLine`
  * `importDeclaration.forceMultiLine`
* Detecting AST specific properties differs between the playground and the website:
  * This PR unifies the logic
  * We now regard more properties as AST specific by assuming properties of form `a.b` are an AST specific property of `a` if `a` exists.
    * This leads to properties like `arguments.preferHanging` being (correctly) regarded as AST specific property of `preferHanging`
    * Currently this logic aligns well with the current configuration, but can we assume this will still be the case in the future?
  * I am not too sure about how I display AST specific properties with a different definition than the parent currently. Will be happy about feedback here!

### DIFF
--- a/website/playground/src/plugins/getPluginDefaultConfig.ts
+++ b/website/playground/src/plugins/getPluginDefaultConfig.ts
@@ -15,18 +15,19 @@ export async function getPluginDefaultConfig(configSchemaUrl: string, signal: Ab
       if (propertyName === "$schema" || propertyName === "deno" || propertyName === "locked") {
         continue;
       }
+      const property = json.properties[propertyName];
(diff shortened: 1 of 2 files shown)


dsherret (approved):
Thanks!

PR #832, opened 2024-03-17
feat: gitignore support

### LINKED ISSUE #378: First-class support for .gitignore and .gitattributes
In our project, we have a generator that takes a repository state and reads all `.gitignore` and `.gitattributes` files. For `.gitattributes` specifically we're interested in code that is marked autogenerated (see https://github.com/github/linguist/blob/master/docs/overrides.md#generated-code, basically the `linguist-generated` and `linguist-vendored` flags).
It will effectively produce a `dprint.json` file with the `excludes` and `includes` arrays set that extends a hand-written `dprint.base.json` (on a side note: we can't extend the other way round because `includes` and `excludes` are not inherited between configs).

One thing that would be really cool would be if we could
* teach dprint to respect `.gitignore` files as a first-class citizen for `excludes`.
* teach dprint to respect `linguist-generated` and `linguist-vendored` flags from `.gitattributes` as a first-class citizen for `includes` (`includes`, because each of these flags can override to `=false` in a more specific glob pattern)

Thoughts?

### LINKED ISSUE #833: dprint attempts to format files in .git
Version: 0.45.0

```console
$ mkdir wat
$ cd wat
$ git init
$ dprint init
$ git commit -am "init"
$ git switch -c test.d.ts
$ dprint fmt
dprint fmt
Error formatting /home/jabaile/wat/.git/refs/heads/test.d.ts. Message: Identifier cannot follow number at file:///home/jabaile/wat/.git/refs/heads/test.d.ts:1:3

  20ba1a6a14151f3a90d7638555294cfad31f451f
    ~
Error formatting /home/jabaile/wat/.git/logs/refs/heads/test.d.ts. Message: Identifier cannot follow number at file:///home/jabaile/wat/.git/logs/refs/heads/test.d.ts:1:44

  0000000000000000000000000000000000000000 20ba1a6a14151f3a90d7638555294cfad31f451...
                                             ~
Had 2 errors formatting.
```

This just surprised someone on my team; it doesn't seem like formatting in `.git` will ever be a good idea, even worse than `node_modules`; can this directory be ignored by default?

dsherret:
Takes the .gitignore into account when figuring out what to format. Files that are excluded via the .gitignore can be unexcluded via a negated exclude pattern.

Breaking:

- Negated exclude patterns are no longer inverted since they are used to unexclude an excluded gitignore pattern (but nobody should have been writing negated exclude patterns anyway... I probably just inverted them originally because I was lazy).

Closes #378
Closes #833


dsherret:
FYI @bradzacher, if you have the time, it would be interesting to know how this impacts Canva's codebase and if there's anything that should change in this PR to make things easier. I'm sure there's probably some edge case that this doesn't take into account.

bradzacher:
What's the best way to grab a binary for this PR? Checkout and build?

dsherret:
@bradzacher oh, sorry! Yeah, checkout and build.

```
git clone https://github.com/dprint/dprint
gh pr checkout 832
cargo build # creates ./target/debug/dprint
```

dsherret:
I'll probably merge this one over the weekend.

bradzacher:
Yeah I gotta figure out how to test this out cos we currently have everything wired into nix as a binary deployment setup.

So replacing the dprint binary without a github release url will take a sec.
I'll have time to take a look today.

bradzacher:
Testing the two binaries:
- dprint 0.45.0
    - `output-file-paths | wc -l` 79789
- this branch
    - `output-file-paths | wc -l` 79797

It looks like there are some incorrectly included files due to this behaviour.
For example we have:
```json
{
  "includes": [
	"*.json",
    "!some/path/to/**/*.json",
  ],
}
```
And then
```gitignore
# some/path/to/a/file/.gitignore
!content.json
```

Currently `some/path/to/a/file/content.json` is ignored by dprint.
With this change it will no longer be ignored.

dsherret:
Thanks! Will investigate soon.

bradzacher:
One final example.
We have:
```gitignore
# root .gitignore

**/target

# some/folder/.gitignore

# don't exclude this target folder
!/target
```

And then
```json
{
  "includes": [
    "!web/src/ui/publish/menu/publish_app/target",
  ]
}
```
Those files under `/target` are now included --- which is technically the correct behaviour -- this is a bug in our config generation logic!

dsherret:
> Currently some/path/to/a/file/content.json is ignored by dprint.
> With this change it will no longer be ignored.

Hmmm... tried this one out and it seemed to work. Maybe I'm doing something wrong in this test case: https://github.com/dprint/dprint/pull/832/commits/daf4c8a02864c641b173a8f3072f2b2e90f9c0b2#diff-38287cc5b9dd24e7cd7e8e42fec4f0ece050cef16b7bcc5344138143f4e95413R493

Maybe the real world example is slightly different?

> Another example would be that we have:

For this one, I'm going to go with the new behaviour because it makes unignoring gitignored files possible. Previously dprint would treat negative globs the same as positive globs in "excludes" (so it would just take the `!` and remove it 😅). Now negative globs can be used as a way to unexclude gitignored files.

joscha:
> The logic itself is pretty simple - it's just that it does generate a pretty cooked config file which has includes, excludes, not includes and not excludes.

😬 I feel compelled to say sorry at this juncture XD

bradzacher:
Cycling back here.

A few weeks ago I was able to update our config setup to remove a whole bunch of lines. TL;DR - our Java team decided to stop using google-java-formatter (via dprint) and go back to intellij-format. So that meant there was a bunch of folder opt-ins and opt-outs that I could remove.

That change cut ~200 lines of `includes` (and `!includes`) from the `dprint.json` - ~20% shorter.

Last week I updated Canva to dprint 0.47.1.
Interesting to note that that upgrade in of itself made dprint match some more files - but all the newly matched files were supposed to be matched (i.e. were accidentally excluded before).

Today I updated our `dprint.json` generator to stop considering `.gitignore` files. This resulted in ~600 lines of `excludes` being removed -- our config is now another 65% shorter! It also caused dprint to match another 16 files that (as before) were previously accidentally excluded!

So big win all round on the config front!

---

I'm guessing all of this config simplification will also help us with issues like this one: https://github.com/dprint/dprint/issues/877

PR #839, opened 2024-04-07
perf(npm): use direct binary for startup boost

dalisoft:
This changes/PR should improve startup performance of CLI by up to 5 times and very huge impact for `lint-staged`, `pre-commit` and `pre-push` commits

---

This PR conflicts with #840 as this PR uses different approach to use binary

**Use this or #840**
Approve and merge whichever ways you want/like

Added support for **arm64** on **npm** `build.ts`


dalisoft:
@dsherret Due of unable to test it locally, can you test please this on your machine and report performance improvements?

Thank you

deployment/npm/build.ts
```diff
@@ -86,16 +87,17 @@ if (version == null) {
     "homepage": "https://github.com/dprint/dprint#readme",
     // for yarn berry (https://github.com/dprint/dprint/issues/686)
     "preferUnplugged": true,
+    "engines": {
+      "node": ">=18.19.1"
+    },
     "scripts": {
-      "postinstall": "node ./install.js",
+      "postinstall": "node ./postinstall.js",
```
dsherret, line 94: I think this now requires the post install step? Previously dprint would work even when someone did `npm install --ignore-scripts`.

dalisoft: Yes, it is true. This is cost for performance improvement

dalisoft: @dsherret Or try #840. #840 should work with ignoring scripts

Issue #841, opened 2024-04-09
`associations` do not overwrite file extensions supported by the plugin.


zhylmzr:
If I override the extension supported by the plugin, then the formatting ignores the files supported by the plugin, which is confusing.

Usually plugins only support one or two extensions, but `prettier` supports dozens of file extensions, so if I want to add a new file extension to `prettier` I have to manually type in all the extensions supported by `prettier`.

```jsonc
{
  "prettier": {
    "associations": [/** all prettier support file extension glob */, "**/*.new_html_extension"],
    "new_html_extension.parser": "html"
  }
}
```

ghuh:
It would be really nice if there was an `append_associtations` or something like that

PR #842, opened 2024-04-09
feat: `association` option don't overwrite plugin supports

zhylmzr:
fixed: https://github.com/dprint/dprint/issues/841

### DIFF
--- a/crates/dprint/src/paths.rs
+++ b/crates/dprint/src/paths.rs
@@ -128,19 +128,15 @@ fn get_plugin_patterns<'a>(plugins: impl Iterator<Item = &'a PluginWithConfig>) 
   let mut file_exts = HashSet::new();
   let mut association_globs = Vec::new();
   for plugin in plugins {
-    let mut had_positive_association = false;
     if let Some(associations) = plugin.associations.as_ref() {
       for pattern in process_config_patterns(associations) {
         if !is_negated_glob(&pattern) {
-          had_positive_association = true;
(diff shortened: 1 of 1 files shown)


crates/dprint/src/paths.rs
```diff
@@ -128,19 +128,15 @@ fn get_plugin_patterns<'a>(plugins: impl Iterator<Item = &'a PluginWithConfig>)
...
           association_globs.push(pattern);
         }
       }
     }
-    if !had_positive_association {
-      file_names.extend(&plugin.file_matching.file_names);
-      file_exts.extend(&plugin.file_matching.file_extensions);
-    }
+    file_names.extend(&plugin.file_matching.file_names);
+    file_exts.extend(&plugin.file_matching.file_extensions);
```
dsherret, line 139: What extension do you need for the prettier plugin that it currently doesn't support? I'm not sure about this change because it will certainly break someone.

zhylmzr: Tencent's miniprogram has a `.wxml` extension file, which is similar to HTML. In Prettier, i can configure its parser through overrides.  https://github.com/prettier/prettier/pull/9081#issuecomment-681927225

This change could indeed have unexpected impacts on existing configurations. A better approach would be to modify the Prettier plugin to support returning the configured `*.parser` association.

Currently, I support it by manually configuring all the files that need to be formatted using Prettier, so I will close this.

PR #868, opened 2024-06-28
feat: implement --staged flag

todor-a:
This is an attempt to implement/port biome's `--staged` [flag](https://biomejs.dev/reference/cli/#:~:text=to%20file%20system.-,%2D%2Dstaged,-%E2%80%94%20When%20set%20to) when formatting/checking.

The implementation just takes over the `arg_includes` field of the glob pattern field which I am not too sure about. I was wondering if it should be a dedicated property but at the end went with just reusing `arg_includes`. Which poses the question, should I add validation that `--staged` and specific files param can't be used together?

let me know what do you think when you get a chance @dsherret


crates/dprint/src/paths.rs
```diff
@@ -91,12 +91,24 @@ pub async fn get_and_resolve_file_paths<'a>(
...
+  }
+
   if file_patterns.config_includes.is_none() {
     // If no includes patterns were specified, derive one from the list of plugins
     // as this is a massive performance improvement, because it collects less file
     // paths to examine and match to plugins later.
     file_patterns.config_includes = Some(GlobPattern::new_vec(get_plugin_patterns(plugins), cwd.clone()));
   }
+
+  // if args.allow_node_modules
```
dsherret, line 111:
```suggestion
```

crates/dprint/src/paths.rs
```diff
@@ -91,12 +91,24 @@ pub async fn get_and_resolve_file_paths<'a>(
 ) -> Result<GlobOutput> {
   let cwd = environment.cwd();
   let mut file_patterns = get_all_file_patterns(config, args, &cwd);
+
+  if args.only_staged {
+    if let Ok(staged_files) = environment.get_staged_files() {
```
dsherret, line 96: I think we should surface the error? Otherwise people might be confused about the behaviour when this fails.

dsherret (commented):
I think we should surface the error on failure, but looks great otherwise.

dsherret (approved):
Thanks! I'm going to do a release this evening so I pushed some changes to get this merged.

PR #872, opened 2024-06-30
chore: use malva and markup_fmt plugins internally

dsherret:
FYI @g-plane - Thanks!

### DIFF
--- a/website/playground/src/external/react-splitpane.css
+++ b/website/playground/src/external/react-splitpane.css
@@ -9,10 +9,12 @@
   -webkit-background-clip: padding;
   background-clip: padding-box;
 }
+
 .Resizer:hover {
   -webkit-transition: all 2s ease;
   transition: all 2s ease;
 }
+
 .Resizer.horizontal {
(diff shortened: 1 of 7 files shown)


website/playground/src/external/react-splitpane.css
```diff
@@ -9,12 +9,10 @@
   -webkit-background-clip: padding;
   background-clip: padding-box;
 }
-
```
dsherret, line 12: @g-plane this feels like a bug?

dsherret: Oh, I see that this occurred because the file happened to be UTF-8 w/ BOM (dprint used to maintain the BOM, but now it passes it to the plugins as-is). I manually added back the line breaks.

g-plane:
It looks markup_fmt didn't detect frontmatter so it breaks the templates. I will fix it later.

dsherret:
@g-plane yeah, i'm not sure what's happening there. It only seems to occur when it formats those files.

Issue #879, opened 2024-07-03
`dprint fmt --staged` should not warn when it can't find any files


dsherret:
Using `--staged` should implicitly enable `--allow-no-files`:

```
> dprint fmt --staged
No files found to format with the specified plugins at V:\dprint. You may want to try using `dprint output-file-paths` to see which files it's finding or run with `--allow-no-files`.
```

FYI @todor-a, in case you're interested in fixing this one.

todor-a:
yep, will take a look tonigh/tomorrow

PR #880, opened 2024-07-05
fix(staged): do not error when `--staged` can't find any files

todor-a:
fixes https://github.com/dprint/dprint/issues/879

### DIFF
--- a/website/src/cli.md
+++ b/website/src/cli.md
@@ -52,6 +52,8 @@ To format only files that are staged use the `--staged` flag:
 dprint fmt --staged
 ```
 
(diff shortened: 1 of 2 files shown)


website/src/cli.md
```diff
@@ -52,6 +52,8 @@ To format only files that are staged use the `--staged` flag:
 dprint fmt --staged
 ```
 
+Implicitly enables the `--allow-no-files` flag.
+
```
dsherret, line 56:
```suggestion
```

dsherret (approved):
Thanks!

PR #889, opened 2024-07-11
chore(website): add third-party plugins to playground

### LINKED ISSUE #883: Add pretty_yaml, malva, markup_fmt to playground and `dprint init`
We'll need to update this file first:

https://github.com/dprint/plugins/blob/main/info.json

Then update the website/playground folder.

FYI @g-plane in case you're interested in taking this one on.

g-plane:
Close #883

### DIFF
--- a/website/src/plugins/malva.md
+++ b/website/src/plugins/malva.md
@@ -25,7 +25,7 @@ dprint config add g-plane/malva
 
 This will update your config file to have an entry for the plugin. Then optionally specify a `"malva"` property to add configuration:
 
-```jsonc
+```json
 {
   "malva": {
     // malva config goes here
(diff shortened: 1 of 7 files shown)


website/src/plugins/malva.md
```diff
@@ -25,7 +25,7 @@ dprint config add g-plane/malva
 
 This will update your config file to have an entry for the plugin. Then optionally specify a `"malva"` property to add configuration:
 
-```jsonc
+```json
```
dsherret, line 28: I think these need to be jsonc because of the comment below? I forget if the website handles it right or not

g-plane: If we set it `jsonc`, syntax highlighting and version replacement will be broken.

dsherret (approved):
LGTM. Exciting. Thanks!

Issue #893, opened 2024-07-20
[Meta] Logo Suggestions


g-plane:
I made logos with different colors and let you choose. In all the logos below, the left part stands for "d" and the right part stands for "p".

### No Shadow

32x32:
![a-no_shadow_图层 1_-32x32](https://github.com/user-attachments/assets/75dd9b84-a9dd-45cc-b590-165fdcba1d95)
64x64:
![a-no_shadow_图层 1_-64x64](https://github.com/user-attachments/assets/790c3ba9-3707-4311-b848-66d156605f81)
256x256:
![a-no_shadow_图层 1_-256x256](https://github.com/user-attachments/assets/51877f8c-4e9c-4ff1-b928-36ac3b8abdaa)
512x512:
![a-no_shadow_图层 1_-512x512](https://github.com/user-attachments/assets/50d556fb-5af4-40d5-925b-03d5b1042cd5)

### Shadowed

32x32:
![a-shadowed_图层 1_-32x32](https://github.com/user-attachments/assets/1e8010d7-86db-47f9-9d5b-b362a23af86e)
64x64:
![a-shadowed_图层 1_-64x64](https://github.com/user-attachments/assets/3a9089bc-eded-4688-b72e-7e3d6e7031d5)
256x256:
![a-shadowed_图层 1_-256x256](https://github.com/user-attachments/assets/a02492e1-6c91-4ea0-8d45-d325914c616e)
512x512:
![a-shadowed_图层 1_-512x512](https://github.com/user-attachments/assets/47b9e3b9-e94d-4667-89ed-783ab847a4b7)

### No Shadow

32x32:
![b-no_shadow_图层 1_-32x32](https://github.com/user-attachments/assets/76edfcab-c8c4-415b-b41c-f954cbf7e2b3)
64x64:
![b-no_shadow_图层 1_-64x64](https://github.com/user-attachments/assets/a2ed90a9-993d-41d9-8baf-05afb47d1303)
256x256:
![b-no_shadow_图层 1_-256x256](https://github.com/user-attachments/assets/7eea1dcb-2132-4dae-b84e-41e1d6cb437d)
512x512:
![b-no_shadow_图层 1_-512x512](https://github.com/user-attachments/assets/9545be98-100f-490d-af55-fc424f125249)

### Shadowed

32x32:
![b-shadowed_图层 1_-32x32](https://github.com/user-attachments/assets/0e208756-c655-46aa-8e8f-ff6d280c185a)
64x64:
[...]

PR #894, opened 2024-07-22
docs: support Angular

g-plane:
Since v0.11.0, markup_fmt supports Angular. This PR updates documentation pages.

---

Off topic: I made some logos for dprint in #893 . Can you check them? If you don't like them, I will close that issue.


dsherret (approved):
Thanks!

PR #897, opened 2024-07-30
feat: allow to specify custom config type name

g-plane:
Currently `Configuration` was hard-coded, which can confuse people who is developing plugins. In my plugins, configuration type isn't called `Configuration`, so I have to add some code like this:

- markup_fmt: https://github.com/g-plane/markup_fmt/blob/2f83a808571d9c6ed55b5d00ccc93c923ce9d1e5/dprint_plugin/src/lib.rs#L18
- Malva: https://github.com/g-plane/malva/blob/b4d2471a99697a51710264148a54b025c5944fa9/dprint_plugin/src/lib.rs#L15
- Pretty YAML: https://github.com/g-plane/pretty_yaml/blob/118bad7b3d677e38d68f80f36660b4cb41cbe2aa/dprint_plugin/src/lib.rs#L15

After this PR, I can remove all the code above and update the `generate_plugin_code!` call like this:

```diff
- generate_plugin_code!(PrettyYamlPluginHandler, PrettyYamlPluginHandler);
+ generate_plugin_code!(PrettyYamlPluginHandler, PrettyYamlPluginHandler, FormatOptions);
```

### DIFF
--- a/crates/core/src/plugins/wasm/mod.rs
+++ b/crates/core/src/plugins/wasm/mod.rs
@@ -62,6 +62,9 @@ pub mod macros {
   #[macro_export]
   macro_rules! generate_plugin_code {
     ($wasm_plugin_struct:ident, $wasm_plugin_creation:expr) => {
+      generate_plugin_code!($wasm_plugin_struct, $wasm_plugin_creation, Configuration)
+    };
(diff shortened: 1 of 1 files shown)


dsherret (approved):
Thanks! A lot of this code was done when I was first learning rust so it's not the best 😄

PR #899, opened 2024-08-06
fix: add `allowTrailingCommas: true` to dprint.json and dprint.jsonc schemas

davidbludlow:
This fixes part of Issue https://github.com/dprint/dprint-vscode/issues/94 . (Or so I assume. I haven't tested that.) The other part of that issue will be fixed by PR https://github.com/dprint/dprint-vscode/pull/95 when that is merged and deployed.

This changes https://dprint.dev/schemas/v0.json , which the dprint vscode plugin uses in its calculation of the json schema.


dsherret (approved):
Thanks!

PR #909, opened 2024-08-26
fix(website): add "very experimental" to official plugins only

g-plane:
Currently "Very Experimental" will be added to `preferSingleLine` in all plugins. This PR limits this to official plugins only.


dsherret (approved):
👍

PR #910, opened 2024-08-26
fix(website): fix "pretty_graphql" in playground URL

g-plane:
This PR fixes broken URL with Pretty GraphQL in playground.


dsherret (approved):
Thanks!

PR #911, opened 2024-08-26
Parse playground config with JSON5 instead of JSON

jakebailey:
Every time I go to make a playground link for a bug report, I get confused and then realize that I copied my own config into the editor with comments, which then silently fails.

This PR instead uses `JSON5.parse`. Technically, dprint supports JSONC (so, maybe `jsonc-parser` would be better?), but I honestly don't know that the feature sets are any different and JSON5 was already a transitive dep with an API more similar to `JSON.parse`.

### DIFF
--- a/website/playground/package-lock.json
+++ b/website/playground/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "@dprint/formatter": "^0.4.1",
         "allotment": "^1.20.2",
+        "json5": "^2.2.3",
         "lz-string": "^1.4.4",
         "monaco-editor": "^0.31.1",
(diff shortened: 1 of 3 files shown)


jakebailey:
Unrelated, but the playground site is still on TS 4.5 and 5.6 notes:

image

Not sure if you'd find an overhaul/upgrade PR helpful or not, though.

dsherret (approved):
Thanks!

dsherret:
> Unrelated, but the playground site is still on TS 4.5, and TS 5.6 notes:

Yeah, would be nice to upgrade https://www.npmjs.com/package/react-monaco-editor sometime.

PR #930, opened 2024-10-20
docs: enable all features on docs.rs

mgeisler:
I was reading about [Wasm plugins](https://github.com/dprint/dprint/blob/main/docs/wasm-plugin-development.md) and wondered why I couldn't find the `plugins` module on https://docs.rs/dprint-core. I think it's just because the feature isn't enabled by default.


dsherret (approved):
Thanks!

PR #937, opened 2024-11-17
feat: support resolving paths to home directory

### LINKED ISSUE #934: Give option to extend from relative path from home with ~
I would like to be able to extend from my global dprint config file, which is at `~/.dprint.json`, so that I can add properties to it; however, when I do so, it doesn't recognize the `~` syntax, which means I'm forced to put an absolute path with `/home/my username/.dprint.json`

This is inconvenient, as I want to be able to use the same config files on the multiple machines I use, which don't have the same username, or even the same OS (I also use windows sometimes)

Is there any way you could make the `~` point to $HOME?

Example of usage:
```json
{
    "extends": "/home/<username>/.dprint.json", // would ideally be ~/.dprint.json
    "json": {
        "associations": ["dot_dprint.json.tmpl"]
    }
}
```

itsvyle:
Resolves #934

### DIFF
--- a/crates/dprint/src/utils/resolve_url_or_file_path.rs
+++ b/crates/dprint/src/utils/resolve_url_or_file_path.rs
@@ -101,6 +101,17 @@ pub fn resolve_url_or_file_path_to_path_source(url_or_file_path: &str, base: &Pa
       }
       return Ok(PathSource::new_remote(url));
     }
+  } else if url_or_file_path.starts_with("~/") {
+    // handle home directory
+    match environment.get_home_dir() {
+      Some(home_dir) => {
+        if url_or_file_path.len() < 4 {
+          return Ok(PathSource::new_local(environment.canonicalize(home_dir)?));
+        }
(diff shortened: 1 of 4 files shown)


dsherret (approved):
Thanks! I'll do a follow-up to add tests.

By the way, I recommend not opening PRs from your main branch because then maintainers can't push updates to the PR. Instead it's better to open a PR from a branch. That also makes it so that when there's updates in the dprint repo to main, you can just pull it into your clean `main` branch instead of having to reset it.

crates/dprint/src/utils/resolve_url_or_file_path.rs
```diff
@@ -101,6 +101,17 @@ pub fn resolve_url_or_file_path_to_path_source(url_or_file_path: &str, base: &Pa
       }
       return Ok(PathSource::new_remote(url));
     }
+  } else if url_or_file_path.starts_with("~/") {
+    // handle home directory
+    match environment.get_home_dir() {
+      Some(home_dir) => {
+        if url_or_file_path.len() < 4 {
```
dsherret, line 108: Why is this < 4?

PR #952, opened 2024-12-31
fix(lsp): debug log duration of format request

dsherret:
I've recently had this behaviour with vscode where it takes forever to do a formatting request when also running a cargo build. I logged out this and it seems dprint is responding very quickly < 100ms, so maybe a vscode or dprint-vscode issue.


dsherret:
By the way, my vscode was slow because of GitHub Copilot. Disabling that extension has made my computer fast once again.

Issue #968, opened 2025-03-05
Disable config file searching and only respect one config file in the repo


bradzacher:
In our repo we have a file called `dprint.jsonc` that is NOT a dprint config file. Long story short this is how we configure our binary distribution for dprint.

Recently some infra constraints have been lifted internally which means we no longer need to pass plugins to dprint via CLI flags. This has revealed a problem.

When we run `dprint fmt`, dprint will crawl the entire repo looking for `/dprint.jsonc?/` files. Dprint then finds the aforementioned not-dprint-config-file, loads it and then barfs:
```
No formatting plugins found. Ensure at least one is specified in the 'plugins' array of the configuration file.
```

Previously this error was hidden because we passed `--plugins` via CLI and so it override the lack of a plugins array.

I have tried to work around this by using the `--config` CLI arg -- but this doesn't disable dprint's config file search logic so it still finds the not-config file and errors.

So the request is one of:
- provide a way to ignore a specific file / not treat it as a config
- provide a way to ONLY respect one config file in the repo and ignore all others

The latter would probably be preferable TBH because we expect that each repo has exactly 1 config file in the monorepo root.

adrian-gierakowski:
Would be great to be able to disable config search altogether since in our case we don’t have a dprint.json config file in the repo at all and I guess the search slows down the execution.

dsherret:
Maybe `--no-config-discovery` and `DPRINT_NO_CONFIG_DISCOVERY=1`?

bradzacher:
That would work fine for my usecase!

dsherret:
This is now released in dprint 0.50: https://dprint.dev/cli/#changing-config-discovery

Use this though: `DPRINT_CONFIG_DISCOVERY=false` or `--config-discovery=false` (I decided it would be more usable to add different modes here like `--config-discovery=ignore-descendants`)

PR #982, opened 2025-05-18
feat: add `--config-discovery=false` and `DPRINT_CONFIG_DISCOVERY=false`

dsherret:
Closes https://github.com/dprint/dprint/issues/968


dsherret:
I'm actually thinking to name this: `--config-discovery=disabled` and `DPRINT_CONFIG_DISCOVERY=disabled` because maybe there could be something like `--config-discovery=ancestors` or something

PR #1003, opened 2025-07-02
chore: update deps

chenrui333:
seeing couple of build failures for 0.50.1 release on ventura arm, bumping cargo dependencies seems helpful with it.

relates to https://github.com/Homebrew/homebrew-core/pull/228646


chenrui333:
just merging is fine, thanks @dsherret! ❤️

PR #1009, opened 2025-07-14
feat: run multiple plugins for single file ext or path

### LINKED ISSUE #904: How to make multiple plugins with the same extension work together
I'm trying to make [a plugin](https://github.com/pd4d10/dprint-plugin-package-json) to sort `package.json`, like [sort-package-json](https://www.npmjs.com/package/sort-package-json) does.

While testing locally, it seems to be conflict with [the JSON plugin](https://github.com/dprint/dprint-plugin-json). Only one of them can work when both are added.

Is there any best practice for this case?

g-plane:
Close #904 

I'm not sure how to add tests for this feature.

### DIFF
--- a/crates/dprint/src/plugins/name_resolution.rs
+++ b/crates/dprint/src/plugins/name_resolution.rs
@@ -64,24 +64,22 @@ impl PluginNameResolutionMaps {
       return plugin_names;
     }
 
-    if let Some(file_name) = get_lowercase_file_name(file_path) {
-      if let Some(plugin_names) = self.file_name_to_plugin_names_map.get(&file_name) {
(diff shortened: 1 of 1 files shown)


crates/dprint/src/plugins/name_resolution.rs
dsherret, line 1: I'm not sure about doing this. For example, a common pattern is to have the prettier plugin defined last, but then have other plugins above prettier. I believe this change would cause it to format with the other plugin and then format with prettier?

g-plane: Oh, you're right.

PR #1014, opened 2025-07-30
fix: ignore empty proxy env

### LINKED ISSUE #1013: Plugin download failed (dns error) when `export http_proxy=""`
Version: 0.50.1

How to Reproduce:

Sometimes we can only set an empty value to an env var instead of `unset` it, like what described in this [answer](https://stackoverflow.com/questions/55789409/how-to-unset-env-in-dockerfile).

So imagine if someone do this:
```bash
$ export http_proxy=
$ export https_proxy=

$ dprint -L debug fmt
[DEBUG] Error downloading https://plugins.dprint.dev/g-plane/malva-v0.12.0.wasm (0/2): Error downloading https://plugins.dprint.dev/g-plane/malva-v0.12.0.wasm - Error: https://plugins.dprint.dev/g-plane/malva-v0.12.0.wasm: Dns Failed: resolve dns name ':80': failed to lookup address information: Name or service not known
```
I think we should ignore empty string when reading env var `http_proxy`.

hzy:
Closes #1013.

### DIFF
--- a/crates/dprint/src/utils/url.rs
+++ b/crates/dprint/src/utils/url.rs
@@ -40,7 +40,7 @@ impl ProxyProvider for RealProxyUrlProvider {
       std::env::var(env_var_name.to_uppercase())
         .ok()
         .or_else(|| std::env::var(env_var_name.to_lowercase()).ok())
-        .map(|v| if v.is_empty() { None } else { Some(v) })? // ignore ""
+        .and_then(|v| if v.is_empty() { None } else { Some(v) }) // ignore ""
     }
 
     static HTTP_PROXY: OnceLock<Option<String>> = OnceLock::new();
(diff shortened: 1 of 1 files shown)


crates/dprint/src/utils/url.rs
```diff
@@ -40,6 +40,7 @@ impl ProxyProvider for RealProxyUrlProvider {
       std::env::var(env_var_name.to_uppercase())
         .ok()
         .or_else(|| std::env::var(env_var_name.to_lowercase()).ok())
+        .and_then(|v| if v.is_empty() { None } else { Some(v) }) // ignore ""
```
dsherret, line 43:
```suggestion
        .and_then(|v| if v.is_empty() { None } else { Some(v) })
```

dsherret (approved):
Thanks!

PR #1017, opened 2025-08-19
feat: add loongarch64-unknown-linux support

SkyBird233:
This PR adds basic support for `loongarch64-unknown-linux-[gnu,musl]`.

Note that plugins won't work on loongarch64 yet, as dprint uses wasmer-compiler-cranelift on all targets currently and has no loongarch64 support. However, as wasmer-compiler-llvm supports loongarch64 and is the recommended backend for production according to [wasmer's documentation](https://github.com/wasmerio/wasmer/blob/main/lib/compiler-cranelift/README.md), I would like to switch to this backend for at least loongarch64 but would like to hear about your opinions (on whether we should just switch to wasmer-compiler-llvm outright).

P.S., I have smoke tested dprint with wasmer-compiler-llvm backend on loongarch64, using dprint's repository and most plugins works out of the box - with the exception that the `exec` plugin needs prebuilt binaries and the typescript plugin would crash with an unreachable error, which I will try to investigate further.

### DIFF
--- a/.github/workflows/ci.generate.ts
+++ b/.github/workflows/ci.generate.ts
@@ -57,6 +57,14 @@ const profileDataItems: ProfileData[] = [{
   os: OperatingSystem.Linux,
   target: "riscv64gc-unknown-linux-gnu",
   cross: true,
+}, {
+  os: OperatingSystem.Linux,
+  target: "loongarch64-unknown-linux-gnu",
+  cross: true,
+}, {
+  os: OperatingSystem.Linux,
+  target: "loongarch64-unknown-linux-musl",
+  cross: true,
 }];
 const profiles = profileDataItems.map(profile => {
   return {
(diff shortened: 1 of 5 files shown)


SkyBird233:
The issue related to TypeScript and LLVM compiler has been addressed in [this pull request](https://github.com/wasmerio/wasmer/pull/5707).

SkyBird233:
Hi @dsherret, in case you missed the message, how do you think of migrating to wasmer-compiler-llvm mentioned above? There was some uncertainties at first but all problems have been addressed now.

dsherret:
Oh, yeah I missed that. I think integrating wasmer-compiler-llvm would be good specifically for loongarch64 if cranelift doesn't work there. IIRC I think wasmer-compiler-llvm requires installing llvm separately so the experience might not be great though.

SkyBird233:
That's resonable. What do you think would be a good way to specify a default backend for LoongArch, and what about @heiher's idea on automatically picking the backend or letting user to choose?

SkyBird233:
Oh okay, I'll give it a try and open another PR when ready.

PR #1036, opened 2025-11-27
feat: add native Windows arm release

dsherret:


dsherret:
Going to close this one because it's using patches and I'll have an easier solution once https://github.com/dprint/dprint/pull/1178 has landed.

PR #1085, opened 2026-01-29
chore: generate release artifact attestations

scop:
Ref https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations

Untested, but I suppose it could work. ci.yml generation untested as well.


dsherret:
Oh, I think duplicate of https://github.com/dprint/dprint/pull/1067 (your pr)

PR #1109, opened 2026-03-20
feat: use LLVM backend for LoongArch to support plugins

SkyBird233:
Continuation of #1017.

Currently, dprint itself runs on loongarch64, but it's unable to load any plugins because there's no Cranelift support for loongarch64. This pull request enables wasmer's LLVM backend for loongarch64 and sets up cross-compilation, including LLVM, in CI. The QEMU setup in CI is only used to allow llvm-sys to use a foreign `llvm-config` to detect LLVM.

The binaries were tested with `ghcr.io/loong64/debian:trixie-slim` and `registry.alpinelinux.org/img/alpine:3.21`.

Example commands for building binaries locally:
```
docker build \
    -t dprint-cross-base-loongarch64-unknown-linux-gnu \
    -f cross-loongarch64.Dockerfile \
    --build-arg 'CROSS_BASE_IMAGE=ghcr.io/cross-rs/loongarch64-unknown-linux-gnu:main' \
    --build-arg 'TARGET=loongarch64-unknown-linux-gnu' .

export CROSS_TARGET_LOONGARCH64_UNKNOWN_LINUX_GNU_IMAGE=dprint-cross-base-loongarch64-unknown-linux-gnu
cross build \
    -p dprint \
    --locked \
    --target loongarch64-unknown-linux-gnu
```


dsherret (approved):
Thanks!

PR #1131, opened 2026-05-11
fix(install.sh): use correct riscv64 build name, add loongarch64 support, append .exe to path for Windows and resolve relative DPRINT_INSTALL path

jonasgeiler:
Made a few fixes to the `https://dprint.dev/install.sh` script, namely:
- **Use `riscv64gc` build name instead of the old `riscv64`.**
  This seems to have changed somewhere along the way and gone unnoticed until now. The releases only contain a `dprint-riscv64gc-unknown-linux-gnu.zip` file nowadays and the script would fail on `riscv64` systems.
- **Add support for `loongarch64` architecture.**
  According to my research, `uname -sm` prints `Linux loongarch64` for `loongarch64` systems, and since there is a build file for this architecture I figure we should use it.
- **Append `.exe` file extension to executable on Windows.**
  Previously, the script would fail if it detected Windows because the `dprint-x86_64-pc-windows-msvc.zip` file only has a `dprint.exe` file in it, not a `dprint` file like the other release files.
- **Resolve relative `DPRINT_INSTALL` paths.**
  For example if running with `DPRINT_INSTALL=./.dprint curl -fsSL https://dprint.dev/install.sh | sh` the script would fail while unzipping.

And I also took the time to deduplicate the GNU/musl libc detection.

Hope it's okay I packed multiple fixes into a single PR.
Let me know what you think! :shipit:

### DIFF
--- a/website/src/assets/install.sh
+++ b/website/src/assets/install.sh
@@ -36,18 +36,20 @@ else
 	dprint_uri="https://github.com/dprint/dprint/releases/download/${1}/dprint-${target}.zip"
 fi
 
-dprint_install="$(realpath "${DPRINT_INSTALL:-$HOME/.dprint}")"
+dprint_install="${DPRINT_INSTALL:-$HOME/.dprint}"
 bin_dir="$dprint_install/bin"
+if [ ! -d "$bin_dir" ]; then
+	mkdir -p "$bin_dir"
+fi
+dprint_install="$(realpath "$dprint_install")"
+bin_dir="$dprint_install/bin"
+
 exe="$bin_dir/dprint"
 zip="$exe.zip"
 
 # append .exe for Windows
 if [ "$target" = "x86_64-pc-windows-msvc" ]; then
 	exe="$exe.exe"
-fi
-
-if [ ! -d "$bin_dir" ]; then
-	mkdir -p "$bin_dir"
 fi
 
 # download
--- a/website/src/assets/install.sh
+++ b/website/src/assets/install.sh
@@ -15,28 +15,19 @@ else
 	case $(uname -sm) in
 		"Darwin x86_64") target="x86_64-apple-darwin" ;;
 		"Darwin arm64") target="aarch64-apple-darwin" ;;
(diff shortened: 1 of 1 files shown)


website/src/assets/install.sh
```diff
@@ -15,55 +15,52 @@ else
...
+fi
 
 if [ $# -eq 0 ]; then
 	dprint_uri="https://github.com/dprint/dprint/releases/latest/download/dprint-${target}.zip"
 else
 	dprint_uri="https://github.com/dprint/dprint/releases/download/${1}/dprint-${target}.zip"
 fi
 
-dprint_install="${DPRINT_INSTALL:-$HOME/.dprint}"
+dprint_install="$(realpath "${DPRINT_INSTALL:-$HOME/.dprint}")"
```
dsherret: It looks like this realpath is failing because the directory hasn't been created yet. Is it necessary?

dsherret (commented):
Thanks! It looks good other than the realpath.

jonasgeiler:
> @dsherret  Thanks! It looks good other than the realpath.

Oh I didn't know about this behaviour of realpath... Only tested it on Linux where that doesn't happen. I'll try to find another solution.

website/src/assets/install.sh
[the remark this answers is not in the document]
jonasgeiler: Fixed in 9af8ce1c61307d7d7b497716fd3fafd08364f1e6 using a more efficient solution than this.

jonasgeiler:
@dsherret 
I have fixed it now and even tested it in Alpine Linux to make sure it works.
Hope you're happy with the solution!

website/src/assets/install.sh
jonasgeiler: Fixed in 9af8ce1c61307d7d7b497716fd3fafd08364f1e6.
It is necessary if someone sets `DPRINT_INSTALL` to a relative path.

dsherret (approved):
LGTM. Thanks!

PR #1134, opened 2026-05-16
feat: support resolving plugins via npm specifiers and node resolution

### LINKED ISSUE #998: Support plugins on npm via npm specifiers?
```json
{
  “plugins”: [
    “npm:dprint-plugin-typescript@0.87”
  ]
}
```

dsherret:
Adds support for referencing plugins by npm specifiers in `dprint.json`.

## Supported forms

- `npm:@scope/name@version` — wasm plugin, resolved from the npm registry.
- `npm:@scope/name@version@checksum` — same, with a tarball checksum (required for non-wasm plugins).
- `npm:@scope/name@version/plugin.json` — process plugin (the existing process-plugin manifest format, shipped inside the npm package).
- `npm:@scope/name` — no version, resolved from `node_modules` walking up from the config file (lets you pin via your project's `package-lock.json` instead of in dprint.json).

For process plugins distributed via npm, the platform-specific binary can either be referenced via a relative path inside the same package (registry resolves it from the extracted tarball) or via a sibling `npm:` reference in `plugin.json` (node_modules resolves the platform package from `node_modules` alongside).

## Registry configuration

Registry URL is resolved in this order: `NPM_CONFIG_REGISTRY` env var → `.npmrc` walking up from the config dir (scoped registries supported) → `~/.npmrc` → `https://registry.npmjs.org`. Each `PluginCache` instance memoizes the resolved registry to keep `.npmrc` reads from happening on every cache lookup.

## Caching

- Plugin cache manifest schema bumped 8 → 9. Old entries are auto-cleared.
- npm cache keys include the resolved registry URL and the specifier path, so the same `name@version` from different registries (e.g. private mirror via `.npmrc`) and different plugin files within one package don't collide.
- Versioned npm tarballs are extracted to `<cache>/npm/<registry-host>/<name>@<version>/`. `forget` and `clear-cache` both clean this up.
- For node_modules process plugins, `file_hash` covers `plugin.json` *and* the resolved platform zip, so a reinstall that swaps the zip without touching `plugin.json` busts the cache.

## Tarball safety

- Entries with `..` or other paths that resolve outside the extract dir are rejected.
- All entries must share a single top-level wrapper directory (npm convention); mixed or wrapper-less tarballs are rejected rather than silently dropping files.

Closes https://github.com/dprint/dprint/issues/998


crates/dprint/src/commands/config.rs
```diff
@@ -142,7 +171,18 @@ pub async fn add_plugin_config_file<TEnvironment: Environment>(
   } else {
     let mut urls = Vec::with_capacity(plugin_names_or_urls.len());
     for plugin_name_or_url in plugin_names_or_urls {
-      if let Some(url) = resolve_plugin_url_to_add(plugin_name_or_url, &config_path, &config.plugins, environment, plugin_resolver).await? {
+      if let Some(url) = resolve_plugin_url_to_add(
```
dsherret, line 179: Need to change to a struct argument for the parameters (not the services like environment, plugin_resolver)

crates/dprint/src/commands/config.rs
```diff
@@ -235,6 +303,178 @@ async fn resolve_plugin_url_to_add<TEnvironment: Environment>(
...
+  package_json_additions: &mut Vec<(String, String)>,
+  environment: &TEnvironment,
+) -> Result<String> {
+  let parsed = crate::utils::parse_npm_specifier(text)?;
+  // top-level plugin reference: enforce the same .wasm/.json constraint that
+  // parse_plugin_source_reference applies, since `dprint add` writes the
+  // result straight into the plugins array.
+  crate::utils::validate_plugin_extension(&parsed.specifier, text)?;
+
+  let start_dir = config_path.parent();
```
dsherret: Maybe just error if `None` here? It doesn't make sense for a config path not to have a parent.

crates/dprint/src/commands/config.rs
```diff
@@ -235,6 +303,178 @@ async fn resolve_plugin_url_to_add<TEnvironment: Environment>(
...
+
+/// Writes new `devDependencies` entries into the nearest `package.json`
+/// (walking up from the dprint config). Updates an existing entry in place;
+/// appends new ones. Bails if no `package.json` is anywhere along the walk
+/// — `--package-json` is opt-in, so a missing file is a user error rather
+/// than something we want to paper over by creating a file silently.
+fn apply_package_json_additions<TEnvironment: Environment>(
+  config_path: &CanonicalizedPathBuf,
+  additions: &[(String, String)],
+  environment: &TEnvironment,
```
dsherret: Use `impl Environment` instead of a type parameter.

crates/dprint/src/commands/config.rs
```diff
@@ -235,6 +303,178 @@ async fn resolve_plugin_url_to_add<TEnvironment: Environment>(
...
+      pkg_path = Some(candidate);
+      break;
+    }
+  }
+  let pkg_path = pkg_path.ok_or_else(|| {
+    anyhow!(
+      "--package-json was set but no package.json was found at or above {}. Run `npm init -y` first or drop the flag.",
+      start_dir.display(),
+    )
+  })?;
```
dsherret: Test for this?

crates/dprint/src/commands/config.rs
```diff
@@ -235,6 +303,178 @@ async fn resolve_plugin_url_to_add<TEnvironment: Environment>(
...
+    if additions.len() == 1 { "y" } else { "ies" },
+  );
+  Ok(())
+}
+
+/// Returns true if any `package.json` found walking up from `start_dir` lists
+/// `package_name` under `dependencies` or `devDependencies`. Monorepos
+/// commonly list deps at the workspace root rather than each package, so we
+/// keep climbing past package.jsons that don't mention the plugin.
+fn is_in_package_json_deps<TEnvironment: Environment>(package_name: &str, start_dir: &std::path::Path, environment: &TEnvironment) -> bool {
```
dsherret: Again, `&impl Environment`

crates/dprint/src/commands/config.rs
```diff
@@ -235,6 +303,178 @@ async fn resolve_plugin_url_to_add<TEnvironment: Environment>(
...
+  use jsonc_parser::parse_to_value;
+
+  for dir in start_dir.ancestors() {
+    let pkg_path = dir.join("package.json");
+    let Ok(text) = environment.read_file(&pkg_path) else {
+      continue;
+    };
+    let Ok(Some(JsonValue::Object(obj))) = parse_to_value(&text, &Default::default()) else {
+      // malformed package.json — don't pretend we know either way; keep
+      // walking in case a parent package.json is well-formed and lists it.
```
dsherret: Maybe warn in this case instead?

crates/dprint/src/commands/config.rs
```diff
@@ -497,6 +737,55 @@ async fn get_plugins_to_update<TEnvironment: Environment>(
...
     };
 
+    // npm specifiers update via the npm registry (dist-tags.latest), not the
+    // plugin's update_url which would migrate us off the npm form
+    if let PathSource::Npm(npm_source) = &plugin_reference.path_source {
+      if npm_source.specifier.version.is_none() {
+        // unversioned specifiers track node_modules — versions are managed by npm/package-lock
+        log_warn!(
+          environment,
+          "Skipping {} (unversioned npm specifier — update via your package manager).",
```
dsherret, line 829: Test for this if there isn't one?

crates/dprint/src/plugins/implementations/process/setup_process_plugin.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(unix)]
+use anyhow::Context;
```
dsherret: Never use a `cfg` on a `use`. Instead explicitly call the trait method where it's used like `anyhow::Context::<etc>`

crates/dprint/src/plugins/implementations/process/setup_process_plugin.rs
```diff
@@ -40,65 +42,115 @@ fn get_plugin_executable_file_path(dir_path: &Path, plugin_name: &str) -> PathBu
...
-        "Plugin zip file did not contain required executable at: {}",
-        plugin_executable_file_path.display()
-      );
-    }
+async fn setup_from_zip<TEnvironment: Environment>(
+  plugin_cache_dir_path: &Path,
+  plugin_name: String,
+  zip_bytes: &[u8],
+  environment: &TEnvironment,
+) -> Result<SetupPluginResult> {
```
dsherret, line 99: Shouldn't this zip extraction do the extract to a temp dir beside the destination and then rename? It would be more reliable if that's not already done. Maybe `extract_zip` already does it.

crates/dprint/src/plugins/implementations/process/setup_process_plugin.rs
```diff
@@ -40,65 +42,115 @@ fn get_plugin_executable_file_path(dir_path: &Path, plugin_name: &str) -> PathBu
...
+    );
   }
+
+  start_communicator_and_collect_info(plugin_executable_file_path, plugin_name, environment).await
+}
+
+/// Installs a pre-resolved executable into the plugin cache. The bytes come
+/// straight from a per-platform npm package, which ships the binary directly
+/// (no inner zip), so we just write them and chmod +x.
+async fn setup_from_binary<TEnvironment: Environment>(
```
dsherret: This seems very wrong. It should include the entire tarball from npm packages and shouldn't be setting up a single binary.

PR #1155, opened 2026-06-20
feat: support NO_COLOR and FORCE_COLOR env vars

### LINKED ISSUE #391: force colors on CI?
On [Buildkite](https://buildkite.com/) the diff coloring from `dprint check` does not seem to be applied (neither seems to be https://github.com/dprint/dprint/blob/f33209b9192ba984d125f7a08a16731eddb0145a/crates/dprint/src/cli/run_cli.rs#L291).

I tried (unsuccessfully):
* `TERM=xterm-256color` (Buildkite's default `TERM`) (this seems to be what https://github.com/crossterm-rs/crossterm is using)
* `CLICOLOR=1` (via https://github.com/mitsuhiko/clicolors-control/blob/master/src/common.rs)
* `CLICOLOR_FORCE=1` (via https://github.com/mitsuhiko/clicolors-control/blob/master/src/common.rs)

Thoughts:
* I am holding it wrong (do you have another suggestion?)
* introducing a `--color` option (similar to https://github.com/rust-lang/rustfmt/pull/2137)

dsherret:
Closes https://github.com/dprint/dprint/issues/391

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -917,6 +917,7 @@ dependencies = [
  "console_static_text",
  "crossterm",
  "deno_npmrc",
+ "deno_terminal",
  "dirs",
  "dissimilar",
  "dprint-core",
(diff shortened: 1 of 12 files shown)


joscha:
🎉

PR #1170, opened 2026-06-22
feat: add `appendAssociations` plugin config to extend a plugin's default file matching

### LINKED ISSUE #841: `associations` do not overwrite file extensions supported by the plugin.

dsherret:
Closes #841

### DIFF
--- a/crates/dprint/src/commands/formatting.rs
+++ b/crates/dprint/src/commands/formatting.rs
@@ -1058,6 +1058,36 @@ mod test {
   }
 
   #[test]
+  fn should_format_files_with_append_associations() {
+    let file_path1 = "/file1.txt"; // matched by default extension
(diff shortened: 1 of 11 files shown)


dsherret:
I think I'm going to redesign associations to always be additive and then you can exclude stuff via negated globs. Seems to currently be a footgun and I feel like some people are improperly using it when I do a search for associations in dprint.json files on github.

PR #1178, opened 2026-06-24
feat: migrate wasm plugin runtime from wasmer to wasmtime

dsherret:
This will allow us to support many more architectures (ex. windows arm) and drop a lot of code. I checked the performance and it's equivalent.

--

Replaces the wasmer crate with wasmtime (43.0.2) for compiling and running Wasm format plugins.

- Host state now lives in the wasmtime Store (a WasmHostState enum reached via Caller) instead of wasmer's FunctionEnv; imports are registered on a Linker and memory is accessed via wasmtime::Memory.
- Native Cranelift codegen on x86_64/aarch64/riscv64/s390x. powerpc64, loongarch64 and android instead compile to wasmtime's portable Pulley bytecode (no native backend, and no signal-based traps in the Android sandbox), replacing the wasmi interpreter (which according to Wasmer, was not good and buggy. They've removed it).
- Drops the corosensei dependency and the loongarch64 LLVM cross-image (Pulley is pure Rust via Cranelift), so loongarch64 now builds with plain cross. Removes the obsolete cross-image workflow and Dockerfile.
- Builds wasmtime with a minimal feature set to keep the binary size down.
- Sets max_wasm_stack to 1 MiB and runs each instance on a dedicated larger-stack thread (wasmtime's default overflows on deeply nested files).


crates/dprint/src/plugins/implementations/wasm/plugin.rs
```diff
@@ -258,77 +266,84 @@ impl<TEnvironment: Environment> InitializedWasmPlugin<TEnvironment> {
...
-              if response.send(result).is_err() {
-                break; // disconnected
+    // spawn the wasm instance on a dedicated thread to reduce issues. use an
+    // explicit large stack: wasmtime runs wasm on this native stack and allows
+    // it up to MAX_WASM_STACK_SIZE, so the thread needs room for that plus the
+    // host frames (tokio blocking threads are too small for this on Windows).
+    std::thread::Builder::new()
+      .name(format!("dprint-wasm-{}", self.name))
+      .stack_size(WASM_WORKER_STACK_SIZE)
+      .spawn({
```
dsherret: Blocker

PR #1184, opened 2026-06-26
feat(npm): add `--checksum` flag to `dprint add`

dsherret:
Adds a `--checksum` flag that forces a checksum onto the written plugin entry even for Wasm plugins (process plugins always carry one). Applies to bare URLs, named plugins, interactive selection, and `npm:` specifiers, and conflicts with `--no-version` / `--package-json` since an unversioned entry cannot carry a checksum.

### DIFF
--- a/crates/dprint/src/commands/config.rs
+++ b/crates/dprint/src/commands/config.rs
@@ -3161,6 +3161,68 @@ mod test {
   }
 
   #[tokio::test]
+  async fn npm_add_explicit_path_in_multi_plugin_package_uses_path_kind() {
+    // a package shipping several plugins: an explicit path picks its own file
+    // and derives the kind from the extension, ignoring a sibling plugin.json.
+    use crate::test_helpers::create_test_npm_tarball;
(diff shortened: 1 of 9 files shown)


crates/dprint/src/commands/config.rs
```diff
@@ -347,16 +366,40 @@ async fn resolve_plugin_url_to_add<TEnvironment: Environment>(
...
+  let parsed = crate::utils::parse_checksum_path_or_url(&url);
+  if parsed.checksum.is_some() {
+    return Ok(url);
+  }
+  let lower = parsed.path_or_url.to_lowercase();
+  if !(lower.ends_with(".wasm") || lower.ends_with(".json")) {
+    return Ok(url); // not a plugin file we can checksum
+  }
+  let parsed_url = Url::parse(&parsed.path_or_url)?;
+  let (_, file) = environment.download_file_err_404(&parsed_url, None).await?;
```
dsherret: This should setup the plugin instead of downloading and throwing away.

PR #1197, opened 2026-06-29
perf: avoid unnecessary save point rewinds

### LINKED ISSUE #195: Conditional save point reverting
The printer currently always jumps back to save points. For performance reasons, it shouldn't jump back unless it needs to because then it just wastes time reprinting the same thing.

To implement this, the printer will need to be able to resolve conditions as if the state of the printer were back at the save point.

todor-a:
## Summary

Avoid rewinding to a printer save point when a condition that was unresolved at the save point now resolves to the same false-path output that was already printed.

The printer now re-resolves the condition at the saved writer state and only restores the save point when the resolved branch would change the printed output. This keeps the existing save-point behavior for required rewrites, but skips the duplicate false-path repaint.

Closes #195.

## Benchmark results

Synthetic release benchmark using path-dependent `dprint-core` builds from `origin/main` and this branch. Lower is better.

| Case | Old | New | Delta |
| --- | ---: | ---: | ---: |
| future resolves false, `10000 x 20`, 50 rounds | 8.797 ms | 7.354 ms | ~16.4% faster |
| future resolves false, `5000 x 100`, 20 rounds | 17.641 ms | 16.375 ms | ~7.2% faster |
| future resolves true, rollback still needed, 100 rounds | 6.292 ms | 6.406 ms | ~1.8% slower |

### DIFF
--- a/crates/core/src/formatting/printer.rs
+++ b/crates/core/src/formatting/printer.rs
@@ -382,6 +382,36 @@ impl<'a> Printer<'a> {
     self.skip_moving_next = true;
   }
 
+  fn update_state_to_save_point_if_needed(&mut self, save_point: &'a SavePoint<'a>) -> bool {
+    if self.should_update_state_to_save_point(save_point) {
+      self.update_state_to_save_point(save_point, false);
+      true
+    } else {
+      false
+    }
+  }
+
+  fn should_update_state_to_save_point(&mut self, save_point: &'a SavePoint<'a>) -> bool {
+    let Some(node) = save_point.node else {
+      return true;
+    };
+    let node = unsafe { &*node.get_node() };
+    let PrintItem::Condition(condition) = &node.item else {
+      return true;
+    };
+
+    self.resolving_save_point.replace(save_point);
+    let mut context = ConditionResolverContext::new(self, save_point.writer_state.writer_info(self.writer.indent_width()));
+    let condition_value = condition.resolve(&mut context);
+    self.resolving_save_point.take();
+
+    if condition.is_stored {
+      self.resolved_conditions.insert(condition.unique_id(), condition_value);
+    }
+
+    condition_value == Some(true) && (condition.true_path.is_some() || condition.false_path.is_some())
+  }
+
   #[inline]
   fn handle_signal(&mut self, signal: &Signal) {
     match signal {
@@ -465,7 +495,7 @@ impl<'a> Printer<'a> {
         self.resolved_line_numbers.insert(line_number_id, self.writer.line_number());
         let option_save_point = self.look_ahead_line_number_save_points.remove(&line_number_id);
         if let Some(save_point) = option_save_point {
-          self.update_state_to_save_point(save_point, false);
+          self.update_state_to_save_point_if_needed(save_point);
         }
       }
       Info::ColumnNumber(column_number) => {
@@ -473,7 +503,7 @@ impl<'a> Printer<'a> {
         self.resolved_column_numbers.insert(column_number_id, self.writer.column_number());
         let option_save_point = self.look_ahead_column_number_save_points.remove(&column_number_id);
         if let Some(save_point) = option_save_point {
-          self.update_state_to_save_point(save_point, false);
+          self.update_state_to_save_point_if_needed(save_point);
         }
       }
       Info::IsStartOfLine(is_start_of_line) => {
@@ -481,7 +511,7 @@ impl<'a> Printer<'a> {
         self.resolved_is_start_of_lines.insert(is_start_of_line_id, self.writer.is_start_of_line());
         let option_save_point = self.look_ahead_is_start_of_line_save_points.remove(&is_start_of_line_id);
         if let Some(save_point) = option_save_point {
-          self.update_state_to_save_point(save_point, false);
+          self.update_state_to_save_point_if_needed(save_point);
         }
       }
       Info::IndentLevel(indent_level) => {
@@ -489,7 +519,7 @@ impl<'a> Printer<'a> {
         self.resolved_indent_levels.insert(indent_level_id, self.writer.indent_level());
         let option_save_point = self.look_ahead_indent_level_save_points.remove(&indent_level_id);
         if let Some(save_point) = option_save_point {
-          self.update_state_to_save_point(save_point, false);
+          self.update_state_to_save_point_if_needed(save_point);
         }
       }
       Info::LineStartColumnNumber(line_start_column_number) => {
@@ -499,7 +529,7 @@ impl<'a> Printer<'a> {
           .insert(line_start_column_number_id, self.writer.line_start_column_number());
         let option_save_point = self.look_ahead_line_start_column_number_save_points.remove(&line_start_column_number_id);
         if let Some(save_point) = option_save_point {
-          self.update_state_to_save_point(save_point, false);
+          self.update_state_to_save_point_if_needed(save_point);
         }
       }
       Info::LineStartIndentLevel(line_start_indent_level) => {
@@ -509,7 +539,7 @@ impl<'a> Printer<'a> {
           .insert(line_start_indent_level_id, self.writer.line_start_indent_level());
         let option_save_point = self.look_ahead_line_start_indent_level_save_points.remove(&line_start_indent_level_id);
         if let Some(save_point) = option_save_point {
-          self.update_state_to_save_point(save_point, false);
+          self.update_state_to_save_point_if_needed(save_point);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


dsherret:
I had AI analyze this PR. I think it's somewhat risky to do, but worth it if we can get it right.

# AI Analysis

I ran this branch against real files and dug into why the output changes.
  Summary: the approach targets the right code, but the predicate that decides
  whether to skip the rewind is checking the wrong thing.

  **What happens now**

  Built against dprint-plugin-toml and run over ~4,500 real `.toml` files, 44
  format differently. They all take this shape:

  ```toml
  proposal = ["transforms", "swc_ecma_transforms_proposal",
  "swc_ecma_transforms_classes", "swc_ecma_transforms_macros",
    "swc_macros_common"]
  ```
  
  An array partly on the opening line, then wrapped, with the closing bracket
  left on the last element — the "single-line mode but a newline happened
  anyway" state that `is_multi_line` exists to prevent.

  **Why**
  
  ```rust
  condition_value == Some(true) && (condition.true_path.is_some() ||
  condition.false_path.is_some())
  ```

  This assumes a condition can only influence output through its own
  `true_path`/`false_path`. That doesn't hold for conditions created with
  `create_reference()` and read through `create_resolver()`. `is_multi_line` in
  `gen_separated_values` is one, and it has **no paths at all** — its value is
  consumed by other conditions, so it changes output produced elsewhere in the
  group. The guard reads that as "no effect", so those are exactly the
  conditions it skips.

  Skipping is also unrecoverable. When the condition was unresolved,
  `handle_condition` stored `None`. Skipping the rewind means `handle_condition`
  never runs again for it, so it stays `None` — and
  `handle_condition_reevaluation` requires a previously *resolved* value, so the
  re-evaluation after the values bails out too. Nothing left can correct it.

  **What would have to change**

  1. **Decide on "would the output be the same", not "is this condition's own 
  branch the same".** For a condition whose value is read by others, that means
  knowing whether anything already printed since the save point actually
  observed it. The printer doesn't record that today; it would need to track
  which condition ids were read via `resolved_condition` between taking the save
  point and reaching it. If the condition wasn't observed, skipping is safe; if
  it was, the rewind is required.

  2. **Treat `None` as "must rewind".** `condition_value == Some(true)`
  currently sends "still unresolved" down the skip path, but unresolved means
  there's nothing to conclude.

  3. **Never store `None` when skipping.** `resolved_condition` won't create a
  new look-ahead save point once the id is present in `resolved_conditions`, so
  caching `None` permanently prevents the condition from ever resolving.

  4. **Keep the condition resolvable after a skip.** If the rewind is skipped,
  the resolved value still has to be recorded so the re-evaluation can flip it
  later if it turns out wrong.

  I tried (2), (3) and (4) on their own, and then a version that forces the
  rewind whenever `condition.is_stored`. That last one restores correct output
  and removes the entire speedup — so the gain currently comes from the unsound
  case rather than from something patchable around the edges. (1) is the part
  that would make it real.

  **Is it still worth doing**

  Yes, but not everywhere. I measured the ceiling by patching the printer to
  skip *every* info look-ahead rewind (wrong output, but it bounds the prize):

  | file | plugin | rewinds | normal | all skipped | ceiling |
  | --- | --- | ---: | ---: | ---: | ---: |
  | `lib.dom.d.ts` (1.9 MB) | TypeScript | 24,488 | 73.2 ms | 70.4 ms | 3.9% |
  | `_tokenizer.ts` (93 KB) | TypeScript | 3,128 | 6.93 ms | 5.41 ms | 21.9% |
  | `crypto_test.ts` (89 KB) | TypeScript | 1,033 | 3.17 ms | 2.83 ms | 10.6% |
  | `Cargo.lock` (287 KB) | TOML | 820 | 6.57 ms | 6.60 ms | 0% |

  So there's a real win for TypeScript and effectively none for TOML. That's an
  upper bound — only the rewinds whose output wouldn't change are actually
  skippable — but it's in the same range as the benchmarks in the PR
  description.

  Worth noting that the pathological nesting case has since been addressed from
  a different angle in #1219 and #1220, so what's left here is the general win
  rather than the blowup.
  
  Given the failure mode is silently reformatted files rather than a crash,
  whatever lands should be gated on a diff over a decent corpus with both
  dprint-plugin-typescript and dprint-plugin-toml.

todor-a:
Thanks for digging into this properly — the analysis is right on every point, and the conclusion at the end of it ("the gain comes from the unsound case rather than from something patchable around the edges") is the one that holds up. I went and tried to disprove that part specifically, including building your (1). It doesn't survive. Details below, and I think this should be closed.

## The bug reproduces, and it's worse than 44 files

No plugin needed. `get_is_multi_line_for_hanging` and `get_is_multi_line_for_multi_line` both construct their condition with `true_path: None, false_path: None`, so `condition.true_path.is_some() || condition.false_path.is_some()` is false for them no matter what they resolve to, and the rewind was skipped unconditionally — including when the group had turned out to be multi-line:

| | output |
| --- | --- |
| `main` | `proposal = [` ⏎ `  "transforms"` ⏎ `  "swc_ecma_transforms_proposal"` ⏎ … ⏎ `]` |
| this branch | `proposal = [` ⏎ `"transforms", "swc_ecma_transforms_proposal", …]` |

One caveat on measuring this that cost me a while: `dprint-plugin-typescript` 0.96.1 requires `dprint-core` 0.67.4, so a `[patch.crates-io]` pointing at this repo's `crates/core` (0.69.0) silently does not apply to it — the plugin keeps using the registry core and every hash matches. `dprint-plugin-toml` 0.8.0 asks for 0.69.0 and does get patched, which is why a TOML-only run looks like a TOML-only bug. With the TS plugin actually bound to this branch's core (path dependency), over ~14.3k files from this repo's specs and `example-references` (prettier, oxc), plus vitest, oxc and ~1.1k TOML files:

**1,389 of 14,294 files format differently from `main`**, TypeScript and JavaScript included — not 44, and not TOML-only.

## Four variants, and the tradeoff between them

Each row is a strictly more careful predicate than the one above it. Timing is `dprint-plugin-typescript` format time over a fixed file set, min of 50 runs, builds interleaved (non-interleaved runs gave me sign errors of several percent, so this matters).

| variant | files differing from `main` | TS format time |
| --- | ---: | ---: |
| this PR as-is | 1,389 | +5–7% |
| skip only on a settled `false` (rewind on `true` and on `None`) | 23 | **+3.9%** |
| …plus verifying every reader's conclusion (your (1)) | 2 | −2.3% |
| skip only when nothing can read the condition (`is_stored` ⇒ rewind) | **0** | −0.8% |

It's monotone. Every step toward correct output costs the win, and nothing correct is faster than `main`.

## Why it's structural rather than a tuning problem

Two isolating probes, on top of the `is_stored`-gated version:

- **pay the re-resolve, then always rewind anyway** → +0.7% (i.e. `main`, within noise). The extra `condition.resolve()` is not the cost.
- **skip every unstored rewind with no predicate at all** → +0.3%. Skipping unstored rewinds is worth nothing.

So the whole 5–7% comes from skipping the *stored* rewinds — the `is_multi_line` ones, which are exactly the ones that must not be skipped. A count of skips is misleading here: the `is_stored`-gated version still skips 121,770 rewinds against 81,195 remaining, but those are the cheap ones. I also tried gating the skip attempt on the rewind spanning at least one line, to only pay the predicate when reprinting is expensive — that immediately breaks this PR's own test, because the skippable rewinds are all same-line.

The prize is small in absolute terms. Skipping *every* stored rewind on `lib.dom.d.ts` saves 2ms out of 58ms. Rewinds here are cheap and resolvers are comparatively expensive, so any predicate costing a resolve loses to the rewind it avoids.

## On (1) specifically

Worth noting that plain read-tracking — "was this condition read via `resolved_condition` since the save point" — cannot work, and not for cost reasons. In `gen_separated_values` the read genuinely happens: `is_multi_line_condition` stores `None`, then `multiLineOrHanging` reads that `None` and prints the single-line path. So the read is always inside the window and read-tracking forces the rewind every time, landing back at the `is_stored` row.

What does move the needle is that `multiLineOrHanging` reads through `create_resolver()`, and for an identity resolver `None` and `Some(false)` yield the same branch — so when `is_multi_line` settles on `false`, the reprint really is unnecessary. Capturing that means tracking each reader's *conclusion*, not the read: record every condition that read a still-unresolved value, along with the writer state it read at and the answer it reached, then at skip time set the value to the settled `false`, re-resolve those readers at their own writer states, and compare. If every answer is unchanged the skip is sound by induction — a reader's value is the only channel out of the condition, so unchanged values mean unchanged paths and unchanged downstream reads.

That is implemented and it does work: 23 → 2 remaining diffs. But it lands at −2.3%, because it needs an extra resolve per reader on every stored-condition settle, which is the same order as the reprint it avoids.

The 2 that survive point at a deeper problem: verification re-runs resolvers, and resolvers are not pure. `get_is_multi_line_for_multi_line` mutates its `last_result` `RefCell` on every call, and `evaluate` reads `value_datas`, which generation mutates. So a speculative re-resolve is not guaranteed to answer what a real replay would. Making this exact would mean requiring purity from plugin resolvers across the ecosystem, which is a much bigger contract change than this optimization is worth.

All this being said, I will go ahead and close this.

PR #1202, opened 2026-07-12
feat(core): convert Utf8Error to FormatError

### LINKED ISSUE #1201: Feature Request: Add `impl From<std::str::Utf8Error> for FormatError`
Currently we only have `impl From<std::string::FromUtf8Error> for FormatError`:

https://github.com/dprint/dprint/blob/bd455f16e063cf975ea99c91ca2170627bfe2dd2/crates/core/src/plugins/plugin_handler.rs#L111

but we also need `impl From<std::str::Utf8Error> for FormatError` where [`std::str::Utf8Error`](https://doc.rust-lang.org/std/str/struct.Utf8Error.html) can be returned from [`std::str::from_utf8`](https://doc.rust-lang.org/std/str/fn.from_utf8.html). This is different from `String::from_utf8`.

arturict:
## Summary

- add the requested `From<std::str::Utf8Error>` implementation for `FormatError`
- verify that the original UTF-8 error remains available through `downcast_ref`

Closes #1201.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p dprint-core --features process should_convert_utf8_error_to_format_error`
- `git diff --check`

The focused test was run under Ubuntu because this Windows environment does not
have the MSVC linker installed.

## AI usage

I used an AI coding assistant to help inspect the issue, implement the small
conversion, and run the checks. I reviewed the patch and verified the focused
test locally.

### DIFF
--- a/crates/core/src/plugins/plugin_handler.rs
+++ b/crates/core/src/plugins/plugin_handler.rs
@@ -108,9 +108,24 @@ impl_format_error_from!(
   &str,
   Box<dyn std::error::Error + Send + Sync + 'static>,
   std::io::Error,
+  std::str::Utf8Error,
   std::string::FromUtf8Error,
   CriticalFormatError,
 );
+
+#[cfg(test)]
+mod tests {
+  use super::FormatError;
+
+  #[test]
+  fn should_convert_utf8_error_to_format_error() {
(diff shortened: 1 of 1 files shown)


dsherret (approved):
Thanks!

PR #1204, opened 2026-07-14
fix: format explicit paths outside the working directory

### LINKED ISSUE #1199: Unable to format absolute or relative paths in project directory
Version: 0.55.1 (latest)

Minimal setup:
```
$ mkdir ~/foo && cd ~/foo
$ cat <<EOF >dprint.json
{
  "json": {
    "jsonTrailingCommaFiles": [],
    "trailingCommas": "jsonc"
  },
  "plugins": [
    "https://plugins.dprint.dev/json-0.22.0.wasm"
  ]
}
EOF

$ dprint output-file-paths
/home/ubuntu/foo/dprint.json

$ dprint fmt dprint.json # works
```

Nothing to format when in sub-directory:
```
$ mkdir bar && cd bar
$ dprint fmt ../dprint.json
No files found to format with the specified plugins at /home/ubuntu/foo/bar. You may want to try using `dprint output-file-paths` to see which files it's finding or run with `--allow-no-files`.

$ dprint fmt ~/foo/dprint.json
No files found to format with the specified plugins at /home/ubuntu/foo/bar. You may want to try using `dprint output-file-paths` to see which files it's finding or run with `--allow-no-files`.
```

Error when two directories down:

```
$ mkdir baz && cd baz
$ dprint fmt ../../dprint.json

thread 'tokio-runtime-worker' (31557) panicked at crates/dprint/src/environment/canonicalized_path_buf.rs:55:7:
Cannot join to /home/ubuntu/foo/bar/baz because the path had ../ (../..).
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ychampion:
## Summary

- resolve relative and absolute CLI paths against their nearest ancestor search base
- cover one- and two-level parent paths plus absolute paths

## Why

From a project subdirectory, explicit paths such as `../dprint.json`, `../../dprint.json`, and an absolute project path either reported no files or panicked before traversal. The resolved pattern base now expands only for positive CLI includes, while config include filtering remains intact.

Closes #1199.

## Validation

- `cargo test -p dprint should_format_parent_paths_from_subdirectory -- --nocapture` — passed
- `cargo test -p dprint patterns::test -- --nocapture` — passed
- `cargo check -p dprint` — passed
- `cargo clippy -p dprint --bin dprint` — passed
- `cargo fmt --all -- --check` — passed

### DIFF
--- a/crates/dprint/src/commands/formatting.rs
+++ b/crates/dprint/src/commands/formatting.rs
@@ -1481,6 +1481,28 @@ mod test {
     assert_eq!(environment.read_file(&file_path).unwrap(), "text1_formatted");
   }
 
+  #[cfg(unix)]
+  #[test]
+  fn should_format_parent_paths_from_subdirectory() {
+    for (cwd, file_arg) in [("/sub", "../file.txt"), ("/sub", "/file.txt"), ("/sub/nested", "../../file.txt")] {
+      let file_path = "/file.txt";
+      let environment = TestEnvironmentBuilder::with_remote_wasm_plugin()
+        .with_local_config("/dprint.json", |c| {
+          c.add_remote_wasm_plugin();
+        })
+        .write_file(file_path, "text")
+        .write_file(format!("{cwd}/other.txt"), "other")
+        .set_cwd(cwd)
+        .initialize()
+        .build();
+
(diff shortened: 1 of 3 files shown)


dsherret (approved):
Thanks!

Issue #947, opened 2024-12-09
Unable to format or exclude files which contain curly braces in the path


tobyh-canva:
Version: 0.47.1

Hi, I'm having an issue with dprint where some files in my repository contain curly braces `{{ }}` in the path. This seems to be because dprint tries to expand the curly braces as glob syntax, but often this glob is invalid - even if it were valid, I doubt the file would be processed correctly!

Unfortunately this occurs even if I try to exclude the file via the config - I assume the error happens before exclusions apply.

Is there some way for us to disable path expansion via a CLI arg, or escape the glob syntax? Thanks 🙏 

## Example:

```
dprint fmt -- {{myfile}}.yaml
```

### Expected output:
Nothing, exit code 0

### Actual output:
```
error parsing glob '/{{myfile}}.yaml': nested alternate groups are not allowed
```

tobyh-canva:
Possibly related to #552, #920

Issue #920, opened 2024-09-27
Files to format are not found if the path contains square brackets


paulovieira:
Version: 0.47.2

#### Description: 

The dprint CLI doesn't always work for a path like this: `[a]/file.js`.

#### Steps to reproduce (I'm using linux):

Prepare:

```shell
mkdir temp
cd temp
dprint init # choose the ts/js plugin
mkdir "[a]"
emacs "[a]/file.js"
```

Add some content to the file:

```js
// [a]/file.js
let x = 123
let y = 456
```

Format using the explicit path:

```shell
dprint fmt "[a]/file.js"
No files found to format with the specified plugins at /home/user/temp. You may want to try using `dprint output-file-paths` to see which files it's finding or run with `--allow-no-files`."
```

I got the same error using the escape character: `dprint fmt \[a\]/file.js`.

But with a wildcard worked:

```shell
dprint fmt "*/file.js"
Formatted 1 file.
```

#### Motivation

Some web frameworks use "filesystem-based routes", in which some special characters are interpreted as parameter for routes. This is the case with SvelteKit: https://kit.svelte.dev/docs/routing

Thanks.

adrian-gierakowski:
Dup: https://github.com/dprint/dprint/issues/552

adrian-gierakowski:
There is a way to escape the path to make it work. But a flag to threat path literally would be great

PR #1205, opened 2026-07-20
feat: add `--no-glob` to take args as path instead of glob

### LINKED ISSUE #552: Fails to match svelte route files that contain bracket parameters

trim21:
add `--no-glob` to parse args as file path instead of glob

close https://github.com/dprint/dprint/issues/552 https://github.com/dprint/dprint/issues/920 https://github.com/dprint/dprint/issues/947


dsherret:
I have a change that will make glob-like arguments that match a filepath match the filepath instead of being a glob. I just haven't pushed it yet.

trim21:
ok

PR #1208, opened 2026-07-26
fix: don't duplicate a plugin specified in both a config and its extended config (#1043)

### LINKED ISSUE #1043: Adding `plugins` to an extended config makes the configuration unused
Version: dprint 0.50.2

The following works as excepted:
```jsonc
{
  // https://dprint.dev/config/#extending-a-different-configuration-file
  "extends": "https://raw.githubusercontent.com/BesLogic/shared-configs/92ff30a/.dprint.jsonc",
  "newLineKind": "lf",
  "typescript": {
    // TODO: Temporary configs to reduce changes from our previous prettier configs. Remove these in a separate PR
    "arrowFunction.useParentheses": "preferNone",
    "semiColons": "prefer",
    "operatorPosition": "sameLine",
    "conditionalExpression.operatorPosition": "nextLine",
    "importDeclaration.forceSingleLine": false,
    "module.sortImportDeclarations": "maintain",
    "importDeclaration.sortNamedImports": "maintain",
    "quoteStyle": "preferSingle",
    "jsx.quoteStyle": "preferDouble",
  },
  "malva": {
    // Lowercase in hex is arguably more readable/accessible, this I'm willing to reconsider the current defaults
    "hexCase": "lower",
    // TODO: Temporary configs to reduce changes from our previous prettier configs. Remove these in a separate PR
    "quotes": "alwaysSingle",
    "declarationOrder": null,
  },
  "excludes": [
    // Vendored
    "frontend/public/static/admin",
    "frontend/public/static/rest_framework",
    // Ignore all html files in the backend dir as they'll be django templates.
    "backend/**/*.html",
  ],
}
```

But as soon as I add the top-level key

```jsonc
{
  "plugins": [
    "https://plugins.dprint.dev/typescript-0.95.13.wasm", // also tried with 0.93.0 which is the version from extended configs
  ],
}
```

All the configs (including `typescript` and `excludes` go ignored)

Image

---

Similarly, if I remove `extends` and manually copy the entire configs over, it also works.

hdimer:
Fixes #1043

## Problem

When a config both `extends` another config **and** lists a plugin that the extended config also lists, the plugin ends up duplicated in the resolved plugin list, and the extended configuration is silently ignored.

Minimal repro (from #1043): a local config that extends a shared config and repeats one of its plugins loses the shared `typescript`/`excludes`/etc. settings the moment the local `plugins` array is added.

## Root cause

In `handle_config_file` (`crates/dprint/src/configuration/resolve_config.rs`), the extended config's plugins were appended with:

```rust
resolved_config.plugins.extend(plugins);
```

with no deduplication. So a plugin listed in both configs produced two entries for the same source.

Downstream, in `resolve_plugins_scope`, both entries resolve to the same plugin. The first one `shift_remove`s that plugin's config key out of the shared config map, so the second gets an empty config. `PluginsScope::new` then collects the plugins into an `IndexMap` keyed by plugin *name*, so the second, config-less instance replaces the first — and the extended configuration disappears.

The other two places that combine plugin lists already guard against the duplicate source: the local-config path calls `filter_duplicate_plugin_sources`, and so does `inherit_config`. The `extends` path was the only one missing it.

## Fix

Two parts, because deduplicating by source alone doesn't cover the case in the issue report.

**1. Deduplicate by plugin source when combining `extends` plugins** (`resolve_config.rs`), keeping the higher-precedence (earlier) entry, consistent with the other paths:

```rust
resolved_config.plugins.extend(plugins);
resolved_config.plugins = filter_duplicate_plugin_sources(std::mem::take(&mut resolved_config.plugins));
```

**2. Deduplicate by plugin name once the plugins are resolved** (`resolution.rs`). The config in #1043 pins `typescript-0.95.13.wasm` locally while the extended shared config specifies `0.93.0`. Those are different sources, so step 1 doesn't touch them, yet they resolve to the same plugin and hit exactly the same name collision described above — the extended config's older version even won over the locally pinned one. `resolve_plugins_scope` now keeps the highest precedence plugin per name before associating configuration:

```rust
let plugins = filter_duplicate_plugin_names(plugin_resolver.resolve_plugins(config.plugins.clone()).await?);
```

A plugin's name is only known after it has been resolved, which is why this can't live alongside the source-level dedupe.

## Tests

- `should_dedupe_plugin_specified_in_both_local_and_extended_config` (`resolve_config.rs`) covers the source-level dedupe.
- `should_use_extended_config_when_specifying_same_plugin_as_extended_config` and `should_use_extended_config_when_specifying_different_version_of_extended_config_plugin` (`commands/formatting.rs`) run a real `fmt` and assert the extended config's settings are actually applied, which is the symptom reported in #1043. Each fails without its corresponding part of the fix.

---

Disclosure: the original change was written with AI assistance (Claude Code).

### DIFF
--- a/crates/dprint/src/commands/formatting.rs
+++ b/crates/dprint/src/commands/formatting.rs
@@ -1008,6 +1008,51 @@ mod test {
   }
 
   #[test]
+  fn should_use_extended_config_when_specifying_same_plugin_as_extended_config() {
+    // https://github.com/dprint/dprint/issues/1043
+    let environment = TestEnvironmentBuilder::with_remote_wasm_plugin()
+      .with_remote_config("https://dprint.dev/shared.json", |c| {
+        c.add_remote_wasm_plugin().add_config_section("test-plugin", r#"{ "ending": "shared-ending" }"#);
+      })
+      .with_default_config(|c| {
+        c.add_config_section("extends", r#""https://dprint.dev/shared.json""#).add_remote_wasm_plugin();
+      })
+      .write_file("/file.txt", "text")
+      .build();
+
+    run_test_cli(vec!["fmt", "/file.txt"], &environment).unwrap();
+
+    environment.take_stdout_messages();
+    environment.take_stderr_messages();
+    assert_eq!(environment.read_file("/file.txt").unwrap(), "text_shared-ending");
+  }
+
+  #[test]
+  fn should_use_extended_config_when_specifying_different_version_of_extended_config_plugin() {
+    // https://github.com/dprint/dprint/issues/1043
+    // The local config pins a newer version of a plugin that the config it extends
+    // also specifies. The local version wins and still picks up the extended
+    // config's configuration for it.
+    let environment = TestEnvironmentBuilder::with_remote_wasm_plugin()
+      .add_remote_wasm_0_1_0_plugin()
+      .with_remote_config("https://dprint.dev/shared.json", |c| {
+        c.add_remote_wasm_plugin_0_1_0()
+          .add_config_section("test-plugin", r#"{ "ending": "shared-ending" }"#);
+      })
+      .with_default_config(|c| {
+        c.add_config_section("extends", r#""https://dprint.dev/shared.json""#).add_remote_wasm_plugin();
+      })
+      .write_file("/file.txt", "text")
+      .build();
+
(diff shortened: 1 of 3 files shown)


dsherret (approved):
Thanks!

PR #1217, opened 2026-08-10
feat: add path completion hints to CLI args

### LINKED ISSUE #1216: Tab completion doesn't work
Tab completion does not work in `dprint`. For instance, if I'm in a directory with a single `foo.py` file and I type `dprint fmt f` and press tab, the command will not complete to `dprint fmt foo.py`.

apcamargo:
This PR adds `ValueHint::AnyPath` to CLI arguments that expect a path, so the shell can provide completion hints.

Fixes https://github.com/dprint/dprint/issues/1216


dsherret (approved):
Thanks!

PR #1248, opened 2026-09-09
docs: add Badness, Arity, and Fatou plugins

jolars:
This PR documents the arity, badness, and fatou plugins.


dsherret (approved):
Thanks!

